Issue #1, opened 2018-02-12
Review of current master state (5f0e8ce)


devsnd:
This is a review of the Code in 5f0e8ce (since it's to late to create a PR now).

**General Nitpicks:**

 - Currently you're using [axios](https://github.com/aeternity/aepp-sdk-js/blob/5f0e8ce5e25c11e8fc8ea32faf94a39b5f4b8a6a/lib/aepp-sdk.js#L1) to implement your HTTP interface. On the browser side there is a new standard API for HTTP requests called [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). I think it would be nice to use that module instead, then you had one dependency less on the browser side. (Node would still need to install `fetch-node`)

**Specific Nitpicks:**

 - As long as the HTTP interface is still unstable I think we should not make any presumptions about it. AFAIK it does not yet support SSL. I propose to remove the ["secure" option from the EpochHttpClient](https://github.com/aeternity/aepp-sdk-js/blob/5f0e8ce5e25c11e8fc8ea32faf94a39b5f4b8a6a/lib/aepp-sdk.js#L18) for now.

 - `EpochHttpClient.post` has `options.config` which are actually the headers of the reqeust. It would be nice if it was named `options.headers` instead.

- I think it would be better to allow to let the `getCommitmentHash` parameter `salt` to be filled automatically. Developers might use it incorrectly and we can prevent that by calculating a hash for them. Also in the readme you state that you would use a 256bit int as salt, but such a number does not exist in plain JS. Everything is a 64bit float (and you have integer precision up to 32 bits). So maybe you can ask the core team if having a 32bit salt is secure enough.

- in [`waitForBlock` the `timeout` is actually the polling Interval](https://github.com/aeternity/aepp-sdk-js/blob/5f0e8ce5e25c11e8fc8ea32faf94a39b5f4b8a6a/lib/endpoints/base.js#L11). But maybe a timeout would be a good idea as well.

- it would be nice if made the `example.js` into the actual test-suite of the thing.

**Final Note:**

Your code is nice and tidy and easy to understand. I like it. What's especially interesting is that it shows some inconsistencies in the API that we should collect on Thursday.

[...]

PR #2, opened 2018-02-12
Changes after issue #1 

tillkolter:
@devsnd Raised issue for his code review https://github.com/aeternity/aepp-sdk-js/issues/1

Other than that:
* fixed update pointer problem (wrong pointer key)
* the above fixed the issue with sending token to a name
* cleaned up the example 
* Add license
* Add jsdoc


lib/endpoints/aens.js
```diff
@@ -91,6 +155,29 @@ class AENS {
...
 
+  /**
+   * Executes the complete two-step process for claiming a name
+   *
+   * @param domain
+   * @param preClaimFee
+   * @param claimFee
+   * @returns {Promise<*>}
+   */
+  async fullClaim(domain, preClaimFee, claimFee) {
```
devsnd: I think it would be nice if `salt` was an optional parameter

PR #6, opened 2018-02-14
Move Oracles and websockets logic to SDK

tillkolter:
* Fix Oracle SDK after breaking API changes
* Make oracle reference implementation independent of HTTP interface
* Seperate HTTP and WS logic in seperate provider interfaces


lib/providers/ws/services/oracles.js
```diff
@@ -0,0 +1,202 @@
...
+      'payload': {
+        'type': 'OracleRegisterTxObject',
+        'vsn': 1,
+        'account': this.account,
+        'query_format': queryFormat,
+        'response_format': responseFormat,
+        'query_fee': queryFee,
+        'ttl': {'type': 'delta', 'value': ttl},
+        'fee': fee
+      }
```
devsnd: I feel like in Javascript world, you can make the `register` and `subscribe` calls even prettier by making use of its async nature. In this case you could return a Promise that gets resolved as soon as the registration acknoledgement comes in. This would allow you to write code like this:
```
Oracles.register(...).then(() => Oracles.subscribe())
```
or the even prettier variant:
```
oracle_registration = await Oracles.register(...);
oracle_subscription = await Oracles.subscribe(...);
```

lib/providers/ws/subscriptions.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  update(data) {
+    if ([actions.MINED_BLOCK, actions.NEW_BLOCK].includes(data.action)) {
+      if (this.pendingOracleIds.length > 0) {
+        for (let i in this.pendingOracleIds) {
+          this.emitter.emit('registeredOracle', this.pendingOracleIds[i])
+        }
+        this.pendingOracleIds = []
+      }
```
devsnd: Nitpick:

I feel like you would state your intentions better by popping elements from the `pendingOracleIds` until they are empty. e.g.
```
while (this.pendingOracleIds) {
   const id = this.pendingOracleIds.splice(0, 1);
   emit(...)
}
```

lib/providers/ws/types.js
```diff
@@ -0,0 +1,28 @@
...
+  actions: {
+    MINED_BLOCK: 'mined_block',
+    NEW_BLOCK: 'new_block',
+    REGISTER: 'register',
+    NEW_ORACLE_QUERY: 'new_oracle_query',
+    NEW_ORACLE_RESPONSE: 'new_oracle_response',
+    QUERY: 'query',
+    SUBSCRIBE: 'subscribe'
+  }
+}
```
devsnd: Good idea to formalize these as constants, I'll do the same in the python sdk, who knows whether the core-team will change their minds again about these!

lib/providers/ws/browser/websocket.js
```diff
@@ -0,0 +1,53 @@
...
+      this.readyState = WebSocket.OPEN
+    }
+    this._ws.onerror = (error) => this.emit('error', error)
+    this._ws.onmessage = (message) => this.emit('message', message.data)
+
+  }
+
+  send(message) {
+    this._ws.send(message)
+  }
```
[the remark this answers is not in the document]
tillkolter: Thing is that the Node WebsocketProxy extends the WebSocket class from the npm `ws` package directly, so I would have to override that method as well. But I will consider you recommendation.

lib/providers/ws/services/oracles.js
tillkolter: I agree, that this would be nice from a programmatical perspectiv, but I am not sure, if I should promisify the oracles methods. The websocket implementations I have been introspecting were all event-based.

tillkolter:
@devsnd thx for your input. Putting the provider interface into question is actually underlining my own worries about that design. It was roughly inspired by the way [web3.js](https://github.com/ethereum/web3.js) handles different protocols although the beta documentation is not clear how and if different needs and expectations from the different protocols are unified between the different implementations

lib/providers/ws/subscriptions.js
tillkolter: Good point. I was also thinking about potential concurrency issues (which would cause huge problems anyway atm) and was also considering copying the list before iterating over it.

examples/oracles.js
```diff
@@ -0,0 +1,85 @@
...
+  let client = new AeternityClient(new WebSocketProvider(options.host, options.port))
+
+  let wsProvider = client.provider
+
+  wsProvider.on ('open', function () {
+      console.log ('Websocket connection is open')
+      client.oracles.register ('queryFormat', 'responseFormat', 4, 500, 5)
+    }
+  )
+  wsProvider.on ('registeredOracle', function (oracleId) {
```
dadaphl: I think the example should separate the roles of creators and users of oracles. It's difficult to understand the lifecycle from creating to querying and responding if both roles are mixed in the same place.

muxe:
Is there a reason you "forward" base, aens, accounts, oracles, tx in AeternityClient directly to the provider but not contracts (of http provider)?

PR #10, opened 2018-02-27
Feature/tx creation endpoints

tillkolter:
Add HTTP wrappers and offchain signing functionality for  
* Oracles
* AENS

### DIFF
--- a/lib/providers/http/services/aens.js
+++ b/lib/providers/http/services/aens.js
@@ -66,14 +66,26 @@ class AENS extends HttpService {
    *
    * @param commitment
    * @param fee
+   * @param options
    * @returns {Promise<*>}
    */
-  async preClaim(commitment, fee) {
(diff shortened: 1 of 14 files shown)


lib/providers/http/services/aens.js
```diff
@@ -144,14 +179,26 @@ class AENS extends HttpService {
...
+        account: options && options.account || await this.client.accounts.getPublicKey ()
+      }
+      let {tx} = await this.client.post('tx/name/revoke', payload)
+      return await this.client.tx.sendSigned(tx, privateKey)
+    } else {
+      let {data} = await this.client.post (
+        'name-revoke-tx',
+        payload,
+        {internal: true})
+      return data['name_hash']
```
devsnd: Why would you still use the internal api in this case? e.g. why use `name-revoke-tx` instead of `/tx/name/revoke`?

PR #12, opened 2018-03-21
Feature/call contracts

tillkolter:


lib/providers/http/services/contracts.js
```diff
@@ -47,15 +47,16 @@ class Contracts extends HttpService {
...
+  async getCallTx(contractAddress, callData, options = {}) {
     let payload = {
       ...createTxCallParams(options),
-      'caller': options.caller || await this.client.accounts.getPublicKey(),
+      'caller': options.caller,
       'vm_version': options.vmVersion || 1,
-      'contract': contractPubKey,
-      'function': functionName,
-      'arguments': args.join(',')
+      'call_data': callData,
```
jsnewby: Why lower_case here? Is this key name required by an external API?

lib/providers/http/services/contracts.js
```diff
@@ -70,8 +71,8 @@ class Contracts extends HttpService {
     return data.out
   }
 
-  async encodeCallData(abi, code, func, arg) {
-    let body = {code, abi, 'function': func, arg}
+  async encodeCallData(abi, code, func, args = []) {
+    let body = {code, abi, 'function': func, 'arg': args.join(',')}
```
jsnewby: Why `let` when `body` is never changed?

lib/providers/http/services/contracts.js
tillkolter: yes, its a payload parameter

PR #14, opened 2018-03-26
Release 0.9.0-0.1.0

jsnewby:
Fixes aeternity/aepp-sdk-docs#8


.dir-locals.el
```diff
@@ -0,0 +1,3 @@
+((nil . ((mocha-options . "--recursive --compilers js:@babel/register")
```
tillkolter: speaking of the devil :) this is an emacs file right? i think the repository should be agnostic and this file should be gitignored

.tern-project
```diff
@@ -0,0 +1,7 @@
+{
```
tillkolter: is this obligatory for the project or also more a local file?

dist/aepp-sdk.js
```diff
@@ -0,0 +1,88 @@
+var _classCallCheck = require("@babel/runtime/helpers/classCallCheck");
```
tillkolter: did you make sure to build this right before creating the PR? just want to double check because I would usually forget this :)

lib/providers/http/services/aens.js
```diff
@@ -125,8 +126,10 @@ class AENS extends HttpService {
    * @param fee
    * @returns {Promise<*>}
    */
-  async update(target, nameHash, {nameTtl = 600000, ttl = 1, fee = 1, privateKey, account}) {
-    if (typeof privateKey !== 'undefined') {
+  async update(target, nameHash, account, options = {}) {
```
tillkolter: Might be a seperate later, but we should update the Jdoc when we change the signature.

lib/providers/http/services/base.js
```diff
@@ -73,27 +73,18 @@ class Base extends HttpService {
    * @param recipient
    * @param amount
    * @param fee
```
tillkolter: Update JDoc

dist/aepp-sdk.js
jsnewby: I haven't.

jsnewby: Done.

lib/providers/http/services/aens.js
jsnewby: All done

Issue #21, opened 2018-04-09
params more uniform


muxe:
e.g. 
https://github.com/aeternity/aepp-sdk-js/blob/develop/lib/providers/http/services/aens.js#L128
`async update(target, nameHash, account, options = {}) {`

vs

https://github.com/aeternity/aepp-sdk-js/blob/develop/lib/providers/http/services/aens.js#L193
`async revoke(nameHash, fee = 1, options = {}) {`

in one the account is a param and the fee is in options and in the other the fee is a param and account is in the options

muxe:
solved with #22

PR #22, opened 2018-04-09
Fix/revoke params

muxe:
Since I filed issue #21 I realized it wasn't only the order of params but also strange account and privateKey options.
I basically used the same scheme as in transfer

### DIFF
--- a/lib/providers/http/services/aens.js
+++ b/lib/providers/http/services/aens.js
@@ -194,15 +194,14 @@ class AENS extends HttpService {
     const { fee = 1 } = options
     const { pub, priv } = account
     let payload = {'name_hash': nameHash, fee}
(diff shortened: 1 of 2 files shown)


lib/providers/http/services/aens.js
```diff
@@ -186,21 +186,22 @@ class AENS extends HttpService {
...
    *
    * @param nameHash
-   * @param fee
+   * @param account
    * @param options
    * @returns {Promise<*>}
    */
-  async revoke(nameHash, fee = 1, options = {}) {
+  async revoke(nameHash, account, options = {}) {
+    const { fee = 1 } = options
```
e-user: This can be moved to the function args:
`{ fee = 1 } = {}` instead of `options = {}`

PR #30, opened 2018-05-16
Add standardized change log with current changes from last tag

e-user:
Also capitalized project description to stay in line with upstream.


muxe:
for the re:publica project(s) I changed a lot of æternites to lowercase because we were asked to use lowercase whenever possible. I can't find any official source on how to write æternity though.

PR #33, opened 2018-05-16
Add sign and verify for personal messages

davidyuk:
For the Voting app, it is necessary to implement methods for signing and verifying of personal messages. But I can't find anything that describes how this messages should be signed on Aeternity. I propose to use the approach similar to Bitcoin and Ethereum approaches.
In https://github.com/ethereum/go-ethereum/issues/14794 proposed to use `varint` from Bitcoin to store message length, I suppose to implement it later in case if this whole approach is acceptable.
Depends on #32

### DIFF
--- a/src/utils/crypto.js
+++ b/src/utils/crypto.js
@@ -87,6 +87,21 @@ function verify (str, signature, publicKey) {
   return nacl.sign.detached.verify(new Uint8Array(str), signature, publicKey)
 }
 
+function personalMessageToBinary (message) {
+  const p = Buffer.from('‎Æternity Signed Message:\n', 'utf8')
+  const msg = Buffer.from(message, 'utf8')
+  if (msg.length >= 0xFD) throw new Error('message too long')
+  return Buffer.concat([Buffer.from([p.length]), p, Buffer.from([msg.length]), msg])
+}
+
+function signPersonalMessage (message, privateKey) {
+  return sign(personalMessageToBinary(message), privateKey)
+}
+
+function verifyPersonalMessage (str, signature, publicKey) {
+  return verify(personalMessageToBinary(str), signature, publicKey)
+}
+
 export default {
   hash,
   encodeBase58Check,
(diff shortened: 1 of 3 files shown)


src/utils/crypto.js
```diff
@@ -87,6 +87,19 @@ function verify (str, signature, publicKey) {
   return nacl.sign.detached.verify(new Uint8Array(str), signature, publicKey)
 }
 
+const personalMessageToBinary = (message) => {
```
e-user: Please use `function` instead of assigning anonymous functions to `const`s unless you have a strong reason not to, e.g. because you're assigning a function returned by another function of higher order.

test/crypto.js
```diff
@@ -72,20 +72,53 @@ describe('crypto', () => {
...
-    it('should work together', async () => {
-      let privateKey = Buffer.from(privateKeyAsHex, 'hex')
-      let signature = Crypto.sign(validUnsignedTx, privateKey)
-      let pub = Buffer.from(Crypto.decodeBase58Check(publicKey.split('$')[1]))
-      let verified = Crypto.verify(validUnsignedTx, signature, pub)
-      assert.isTrue(verified)
+
+  describe('sign', () => {
+    it('should produce correct signature', () => {
+      let s = Crypto.sign(txBinary, privateKey)
```
e-user: Use `const` instead of `let` wherever possible (please disregard that old code does it wrong)

test/crypto.js
```diff
@@ -72,20 +72,53 @@ describe('crypto', () => {
...
+
+    describe('verify', () => {
+      it('should verify message', () => {
+        const result = Crypto.verifyPersonalMessage(message, messageSignature, publicKey)
+        assert.isTrue(result)
+      })
+
+      it('should verify message with non-ASCII chars', () => {
+        const result = Crypto.verifyPersonalMessage(
+          messageNonASCII, messageNonASCIISignature, publicKey)
```
e-user: Please remove newline

davidyuk: Don't you have any limits on line length? Usually, I'm trying not to make lines longer than 100 chars.

PR #47, opened 2018-06-08
include examples on how to include the sdk in your project

muxe:
the rest of the docs are still super outdated

PR #51, opened 2018-06-11
include examples on how to include the sdk in your project

muxe:
redid: https://github.com/aeternity/aepp-sdk-js/pull/47


docs/examples.md
```diff
@@ -1,5 +1,92 @@
 # Examples
 
+## Browser &lt;script&gt; tag
```
e-user: Why use HTML quotes in Markdown?

docs/examples.md
```diff
@@ -1,5 +1,92 @@
...
+## nodejs (without compiling)
+
+Use either yarn or npm to install the sdk.
+Use a prebuilt package from npmjs
+```
+yarn add @aeternity/aepp-sdk
+```
+Or use a branch directly from github
+```
+yarn add aeternity/aepp-sdk-js#feature/sweet-api
```
e-user: Should now be `develop`, again

docs/examples.md
```diff
@@ -1,5 +1,92 @@
...
+```
+create index.js
+```js
+const Ae = require('@aeternity/aepp-sdk')
```
e-user: If we compile async/await, we can also assume module syntax can be used, so
`import Ae from '@aeternity/aepp-sdk' seems to be more coherent.

docs/examples.md
```diff
@@ -1,5 +1,92 @@
...
+
+Ae.default.create('https://sdk-testnet.aepps.com').then(client => {
```
e-user: ... which then also means you don't need `.default`, anymore.

PR #52, opened 2018-06-11
Fix issue with URLs and existing path components

### LINKED ISSUE #49: urlparser.resolve breaks path when it comes to test-api
@e-user 

The problem that i'm having is that when i'm defining an absolute path, the library gets only the hostname of the API endpoint and the relative parameters.

Example: `https://test-net-api.aepps.com/api/v2/` -> `https://test-net-api.aepps.com/`

And is using custom strings within the library to check for version and swager, and the entire library becomes unusable.

screen shot 2018-06-09 at 17 10 38

e-user:
Resolves #49.


e-user:
@sadiqevani could you please verify this fix? Thank you!

sadiqevani:
@e-user 

So still not working correctly, let me show you some weird behavior of the URL thingy:

- When the `AETERNITY_EPOCH_API_URL` is `//test-net-api.aepps.com/`
screen shot 2018-06-11 at 21 34 31

- When the `AETERNITY_EPOCH_API_URL` is `//test-net-api.aepps.com/api/` 
screen shot 2018-06-11 at 21 37 44

- When the `AETERNITY_EPOCH_API_URL` is `//test-net-api.aepps.com/api/v2/` 
screen shot 2018-06-11 at 21 38 56

I have no idea what's going on! 

Though the way i would handle API requests would be simply to handle everything internally for the relative URLS, and just ask for the HOST of the node! No specific versioning or extra API.

Later we can use a config object to ask the user which version of the epoch API they want to use, I think this is a recurring pattern on the SDKs.

e-user:
@sadiqevani the right URL to use is `//test-net-api.aepps.com/api/`, because that's what this host (which I'm regretfully not maintaining) decided to expose Epoch through. The request to `/api/api` is completely correct (and not a fatal error!) because starting with Epoch 0.15.0, there is a new endpoint `/api` (without `/v2`!) which results in `/api/api`, given that the poor choice was made to expose Epoch under `/api` in case of this host.  
However all subsequent requests fail because now we're hitting another problem source: The Swagger file reports the base path as `/v2`, and with the prefix slash, this again makes `urlparse.resolve` resolve it as an absolute path.

Please let me quickly issue another fix which I would kindly ask you to test, again.

e-user:
Slightly OT, but I'll address it regardless: 
> Though the way i would handle API requests would be simply to handle everything internally for the relative URLS, and just ask for the HOST of the node! No specific versioning or extra API.

I couldn't disagree more. If you just specify the host of the node, you don't know what protocol to use (at least HTTPS or HTTP), which port - which is already different in case of the host you're using, 80 instead of 3013 - and a base path suffix, which is normally just `/` but in case of this host it's `/api`. So ultimately, you're ending up with the parameters `protocol` (alternatively boolean `secure`), `host`, `port`, `basePath`. Which is basically a terribly inefficient re-implementation of the URL standard. 

> Later we can use a config object to ask the user which version of the epoch API they want to use, I think this is a recurring pattern on the SDKs.

Again I disagree. Ideally, the SDK should automatically do The Right Thing (tm), which is exactly the case with the upcoming release, as it determines the version automatically and picks the right API to choose, whereas with 0.15.0 onwards, the Epoch node tells the SDK how to communicate with it, so it's zero configuration for the user.

e-user:
@sadiqevani would you please try again with my latest changes pushed? And thanks for your patience.

sadiqevani:
> I couldn't disagree more. If you just specify the host of the node, you don't know what protocol to use (at least HTTPS or HTTP), which port - which is already different in case of the host you're using, 80 instead of 3013 - and a base path suffix, which is normally just / but in case of this host it's /api. So ultimately, you're ending up with the parameters protocol (alternatively boolean secure), host, port, basePath. Which is basically a terribly inefficient re-implementation of the URL standard.

> Again I disagree. Ideally, the SDK should automatically do The Right Thing (tm), which is exactly the case with the upcoming release, as it determines the version automatically and picks the right API to use, whereas with 0.15.0 onward, the Epoch node tells the SDK how to communicate with it, so it's zero configuration for the user.

I'm not saying to make it a requirement to provide configuration over the default params, the idea would be that in case the user would want to change something, is capable of doing so, (so, non required configration object to modify versioning and endpoints of the node) if you think of nodes being API endpoints in a distributed way, there might be edge cases where this would be a good idea. But maybe I didn't explain my idea too clearly! 

Working on HapiJS before, I like their mantra which is configuration over code, they give you the "Do the right thing" first, but in case that does not work "Try it in another way" (tm) 😆 

But in this case yeah I can see your point, given that depending on which node you connect too, there might be a different version, and if the node decides which sets of api to use, reduces issues.

sadiqevani:
@e-user checking it out.

sadiqevani:
@e-user 

Just a slight error here: https://github.com/aeternity/aepp-sdk-js/blob/92849edbf032419550b2eda0500db0db3805cfaf/src/client/index.js#L322

by default the property is `undefined`, the regex throws an error there.

e-user:
@sadiqevani now I get where you're going, here. Yes, in theory, it might be useful to leave a backdoor open if someone makes the terrible decision to override Epoch's own hierarchy of paths - I wonder why anyone would ever do this, however.  
I do realize this is something we should absolutely document, also to explain why the SDK makes what requests when initiating the client. To make sure it is at least understood here:  
The SDK just asks for the base URL where Epoch listens, no matter whether there's a reverse proxy in front, or not. It assumes Epoch owns everything under that base URL and queries `/api`, which is the new endpoint for Epoch to provide its own Swagger file it was built with / against. If that doesn't exist (yet), which is the case for < 0.15.0, it falls back to using the information obtained from `/v2/version`, which is already a hack because hypothetically, a newer version of Epoch might not even have `/v2/version` (but maybe `/v3/version`). But this fallback behavior is going away with the next minor release anyway, because we don't want to be carrying around tons of old Swagger files for Epoch versions which no one should be using anymore anyway.  
So the correct behavior is: User provided URL -> request /api (relative to it) -> determine base path -> build full URLs from base URL, base path and path remained per operationId. The only thing required from the user is still protocol, host, port, base path - which is always needed regardless and defaults to what the HTTP(S) standard specifies wrt. ports. If you think there is a real life use case to override this behavior, we should specify this and consider to implement it.

e-user:
@sadiqevani three times a charm, please give it another shot.

sadiqevani:
@e-user 

Good to go!

sadiqevani:
@e-user 

Now which branch should I use? You're merging it into develop, sweet-api still uses the buggy version.

e-user:
@sadiqevani 

feature/sweet-api is already merged into develop, and so will be this branch (i.e. it already contains what used to be in sweet-api).

PR #63, opened 2018-06-27
Don't require Yarn to build the package on "postinstall"

davidyuk:
Alternatively, we can add `yarn` to dev dependencies 🙂


davidyuk:
Original problem: https://github.com/aeternity/hd-wallet-js/pull/1

Issue #9, opened 2018-02-27
Add wallet functionality


tillkolter:
As an SDK user (developer or end-user) I want to be able to import a wallet to be able to sign a transaction and set myself as the sender of certain transactions (`accounts.getPublicKey` would be deprecated then)

e-user:
The new SDK releases provide a reasonable wallet abstraction which wraps an arbitrary number of accounts, all accessed via a common, standardized interface. Spend transactions, in their [default implementation](https://github.com/aeternity/aepp-sdk-js/blob/v0.18.0-0.1.0/es/ae/index.js#L30), use the corresponding public key that belongs to the signing key for the sender of a transaction. It is however easy to set a custom sender, as should be obvious from the code snippet referenced above.

PR #64, opened 2018-06-29
Realworld SDK Refactoring

e-user:
This changeset incorporates changes based on feedback from the Aepps team. It introduces the `stampit` library for composable factories, which means that clients can be built in a more flexible manner.

Documentation updates will be provided in a separate PR / on a separate branch.


e-user:
Fixes #9

PR #65, opened 2018-07-02
Generator of three words identifiers

davidyuk:
Word lists taken from [here](https://github.com/status-im/status-react/tree/develop/src/status_im/utils/gfycat).
https://www.pivotaltracker.com/story/show/157082201


src/utils/identifiers/three-words/index.js
```diff
@@ -0,0 +1,32 @@
...
+ *  PERFORMANCE OF THIS SOFTWARE.
+ */
+
+import adjectives from './adjectives'
+import animals from './animals'
+import { decodeBase58Check } from '../../crypto'
+
+export default address => {
+  const bytes = decodeBase58Check(address.split('$')[1])
+  let entropy = bytes.readUInt32BE(bytes.length - 4)
```
mpowaga: What if this function receives two different addresses with the same entropy? Looks like it would return the same words. Maybe each word should be based on separate entropy (i.e. different bytes of the buffer)?

e-user:
@jsnewby could you please elaborate whether this shall be part of the SDK?

src/utils/identifiers/three-words/index.js
davidyuk: Yes, this function will return the same words if addresses have the same 4 last bytes, but it should be quite difficult for the attacker to generate a key pair with the same last bytes.

Also, this implementation does not protect against typos. Maybe aeternity accounts addresses are already somehow protected against typos?

Entropy can be generated based on the hash of an address or we can use account address as randseed like [Status does](https://github.com/status-im/status-react/blob/dc23ce28039f45204f71da609805fdb7d58f7136/src/status_im/utils/gfycat/core.cljs#L14), but as I understand the generation of such identifier is much cheaper than a generation of key pair so even if we will use the whole address as entropy still will be possible to generate different address with the same identifier, but without corresponding private key. Of course, we can increase the complexity of identifier generation algorithm, but it can decrease the performance of our apps in some cases.

jsnewby:
My opinion is that this ought to be separate from the SDK, as the SDK should be application-agnostic, and this seems to belong to the aepps. Does anyone strongly disagree?

davidyuk:
Three words identifiers are not only for Base aepp. Usefulness of this feature depends on the number of applications that supports it. I think it should be defined as a standard on the level of aepp SDK.

ricricucit:
closed due to inactivity and age. Opened issue #168 in favour of this idea.

davidyuk:
@ricricucit why we can't merge this PR? It is ready to be used in the Base app and other aepps.

es/utils/identifiers/three-words/index.js
```diff
@@ -0,0 +1,32 @@
...
+ *  PERFORMANCE OF THIS SOFTWARE.
+ */
+
+import adjectives from './adjectives'
+import animals from './animals'
+import { assertedType, decodeBase58Check } from '../../crypto'
+
+export default address => {
+  const bytes = decodeBase58Check(assertedType(address, 'ak'))
+  let entropy = bytes.readUInt32BE(bytes.length - 4)
```
davidyuk: Maybe it is more accurate to use `dividedToIntegerBy` and `modulo` from bignumber.js with the whole address instead of working with the last 4 bytes.

nduchak:
Close due to inactivity

PR #75, opened 2018-08-14
Webpack postMessage interference

ricricucit:
This tiny change helps Devs (using ES6 modules) avoid webpack-dev-server postMessage interference.


ricricucit:
Thanks, but I've included this in https://github.com/aeternity/aepp-sdk-js/pull/76

PR #76, opened 2018-08-17
Feature/0.20.0 0.1.0

ricricucit:
New nameHash function as described [here](https://github.com/aeternity/protocol/blob/master/AENS.md#hashing)


ricricucit:
@e-user : I would do this on a release branch when requesting merge to master.......

PR #78, opened 2018-09-03
Implement CLI

nduchak:


bin/utils.js
```diff
@@ -0,0 +1,303 @@
...
+const printError = (msg) => console.log(msg)
+
+const printConfig = ({host}) => {
+  console.log('WALLET_PUB___________' + process.env['WALLET_PUB'])
+  console.log('EPOCH_URL___________' + host)
+}
+
+const printBlock = (block) => {
+  console
+    .log(`Block hash____________________ ${block.hash}
```
e-user: Inconsistency: Here, you are using multiline strings, but below, multiple print statements. Nothing major though.

bin/utils.js
```diff
@@ -0,0 +1,303 @@
...
+Block height__________________ ${block.height}
+State hash____________________ ${block.stateHash}
+Miner_________________________ ${block.miner || 'N/A'} 
+Time__________________________ ${new Date(block.time)}
+Previous block hash___________ ${block.prevHash}
+Transactions__________________ ${block.transactions ? block.transactions.length : 0}`)
```
e-user: For this check as well as all the others where you are using `||` to check for falsy values on the left-hand side as well as `&&` to chain in order to prevent key access to `undefined`: Instead, use `R.path()` (which you already have above!) with either `||` or `R.defaultTo()`, as in (in this case) `R.defaultTo(0, R.path(['transactions', 'length'], block))`. It might look weird at first, but it's more fault-tolerant, works much better for deep cascades of objects and looks more coherent if used everywhere.

bin/utils.js
```diff
@@ -0,0 +1,303 @@
...
+      replace: '*',
+      conform: function () {
+        return true
+      }
+    }
+  }
+}
+
+// FILE I/O
+const writeFile = (name, data) => {
```
e-user: Is there a reason to not use `function` instead, here and below? It's fine if this pattern is used coherently, but throughout most of the rest of the project, `function` is used in preference over lambda assignment, which is all fine by itself otherwise, unless `this` meddling is involved.

es/utils/crypto.js
```diff
@@ -23,10 +23,10 @@
 
 import bs58check from 'bs58check'
 import RLP from 'rlp'
-import {blake2b} from 'blakejs'
+import { blake2b } from 'blakejs'
```
e-user: Is this really necessary? Will you change to `{ X }` style everywhere else in the project so it's at least coherent?

bin/aecli-chain.js
```diff
@@ -0,0 +1,60 @@
...
+//   _____ _           _
+//  / ____| |         (_)
+// | |    | |__   __ _ _ _ __
+// | |    | '_ \ / _` | | '_ \
+// | |____| | | | (_| | | | | |
+//  \_____|_| |_|\__,_|_|_| |_|
+
+const program = require('commander')
+
+require = require('esm')(module/*, options*/) //use to handle es6 import/export
```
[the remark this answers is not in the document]
nduchak: Is it not correct to ESM for require?

package.json
```diff
@@ -5,6 +5,9 @@
   "main": "dist/aepp-sdk.js",
```
ricricucit: Inconsistency: release should match actual release `0.22.0-0.1.0-alpha.2`

es/ae/contract.js
```diff
@@ -69,16 +69,20 @@ async function call (code, abi, address, name, { args = '()', options = {} } = {
 async function deploy (code, abi, {initState = '()', options = {}} = {}) {
   const opt = R.merge(this.Ae.defaults, options)
   const callData = await this.contractEncodeCall(code, abi, 'init', initState)
+  const ownerId = await this.address()
```
ricricucit: clearer and more consistent, thanks.

PR #82, opened 2018-09-25
Feature/postinstall fix

ricricucit:
I've tried installing the lib from a vue app using StandardJS and `postinstall` script was breaking due to `object-curly-spacing`. I've fixed this running `standard --fix`, but I fear a bigger problem exists: our standardJS linter doesn't lint as it should be.


ricricucit:
@nduchak can you please review and then squash-merge (modifying the comment)?

Since it's quite big, I can add that this PR it's about these changes:
- new `webpack` and `webpack-cli`
- standardJS linting using `standard --fix`, mainly to fix `object-curly-spacing` warnings thrown during postinstall task (`npm install`ing the package on a JS app)
- new dev dependencies (for new `wepbpack-cli`)

PR #86, opened 2018-10-02
Add rename package and fix docco comand

nduchak:
Add rename to package.json
Fix docs:docco command


ricricucit:
good fix, but let's hotfix master :)

PR #90, opened 2018-10-08
Remove Left-to-right mark from personal message prefix

davidyuk:
Accidentally, `æternity Signed Message` was copied with the [left-to-right mark](https://en.wikipedia.org/wiki/Left-to-right_mark) (0xe2808e) at the beginning, this completely changes signatures of personal messages without any reason.

Thanks to @CedrikNikita for finding this bug!


davidyuk:
I have no Idea what's wrong with CLI tests.

PR #103, opened 2018-10-17
Fix docker node version

nduchak:
Change docker node version to 10.12.0
Change `Jenkinsfile` to use `pnpm` instead of `npm`


ricricucit:
Thanks @nduchak for merging this, before the big branch :)

PR #115, opened 2018-10-25
Feature/examples fix build

ricricucit:
The example apps were still a bit broken and confused. 
This version should make more clear what an Aepp/Wallet developer should do to start working with `Wallet` and `Aepp` flavors (combined)


ricricucit:
This PR, now has to consider the changes that might come from here: https://github.com/aeternity/aepp-sdk-js/pull/117

PR #116, opened 2018-10-25
Refactoring

davidyuk:


es/utils/async-init.js
```diff
@@ -19,15 +19,14 @@ import stampit from '@stamp/it'
 import * as R from 'ramda'
 
 function asyncInit (options = {}, { stamp, args, instance }) {
-  return R.reduce(async (instance, init) => {
-    instance = await Promise.resolve(instance)
+  return stamp.compose.deepConfiguration.AsyncInit.initializers.reduce(async (instance, init) => {
```
nduchak: Why it is better to use native `reduce` instead of `ramda`.

es/ae/cli.js
```diff
@@ -17,13 +17,13 @@
 
 import Ae from './'
 import Account from '../account/memory'
-import Chain from '../chain/epoch'
+import EpochChain from '../chain/epoch'
```
nduchak: This rename is unnecessary

es/ae/wallet.js
```diff
@@ -27,7 +27,7 @@ import Account from '../account'
 import ContractBase from '../contract'
 import Contract from './contract'
 import Accounts from '../accounts'
-import Chain from '../chain/epoch'
+import EpochChain from '../chain/epoch'
```
nduchak: This rename is unnecessary

ricricucit:
Apart from the changes requested by @nduchak, I'd love to add this to the next release. (0.24.0-0.3.0)

es/ae/cli.js
davidyuk: Not necessary, but it becomes consistent with the import of `EpochContract`. You have base Chain stamp and one connected to Epoch, may be difficult to understand which one is used currently without checking of import path.

es/utils/async-init.js
davidyuk: Let's continue this discussion in https://github.com/aeternity/aepp-sdk-js/pull/120

davidyuk:
@ricricucit I had removed all debatable changes from this PR.

ricricucit:
So should I close this?

ricricucit:
I'm closing this because I've noticed the same changes in https://github.com/aeternity/aepp-sdk-js/pull/120

davidyuk:
I meant that without "async-init: Use native reduce" and "Rename Chain to EpochChain" commits this branch is ready to be merged.

I had duplicate this changes in #120 to avoid conflict resolving. I thought that this branch will get merged faster than #120.

So, I have removed duplicate commits from #120, these PRs are completely independent.

PR #119, opened 2018-10-26
Release 0.24.0-0.2.0 (RPC Improvements + Cli -> Universal)

ricricucit:
### Added
- RPC Client improvement
- (RPC) `onContract` Guard
- (AE PROJECT CLI) born
 ### Changed
- (CLI) New keystore following these specifications: https://www.pivotaltracker.com/n/projects/2124891/stories/155155204
 ### Breaking Change
- the keypair keys changed from `{ pub, priv }` to `{ publicKey, secretKey }` for consistency with other systems using them
 ### Notes and known Issues
- CLI and AE PROJECT CLI will move to a separate package


ricricucit:
@Perseverance can you please check this error? https://ci.aepps.com/blue/organizations/jenkins/aepp-sdk-js/detail/PR-119/4/pipeline

thank you

ricricucit:
unblocked myself, but I'm not sure I solved the conflict in the right way. 
🤔

test/unit/crypto.js
```diff
@@ -37,9 +37,9 @@ describe('crypto', () => {
     it('generates an account key pair', () => {
       const keyPair = Crypto.generateKeyPair()
       assert.ok(keyPair)
-      assert.isTrue(keyPair.pub.startsWith('ak_'))
-      assert.isAtLeast(keyPair.pub.length, 52)
-      assert.isAtMost(keyPair.pub.length, 53)
+      assert.isTrue(keyPair.publicKey.startsWith('ak_'))
+      assert.isAtLeast(keyPair.publicKey.length, 52)
+      assert.isAtMost(keyPair.publicKey.length, 53)
```
davidyuk: Actually, it is an address, not a public key.

ricricucit: we're trying to follow the standard `publicKey/secretKey` that you introduced in hd-walled, for consistency.....or Am I not understanding this comment?

nduchak: I think it's mean that `keyPair.publicKey` have an prefix. And it's look like not a `publicKey`.

davidyuk: It is a terms question, In my opinion, public key encoded into base58check with `ak_` prefix should be named `address` like in Ethereum. `publicKey` should be an array of 32 bytes in accordance with `tweetnacl` and `hd-wallet-js`.
So, I propose to drop the `raw` parameter in `generateKeyPair` and make it return an object with fields `publicKey`, `secretKey`, `address`.

ricricucit: @nduchak @davidyuk I'll go ahead and make a new release (also to help the CLI and CLI PROJECT to move out of this repo), but let's talk about this more in detail, because I don't understand if we did the right thing, or if we have to rename again.

I'd also ask @emil-apeunit and @jsnewby to help with this "naming issue", so we can have some sort of standard across all the systems/apps/integrations we have. Thanks! 🙏

PR #120, opened 2018-10-29
Use ES methods instead of Ramda

davidyuk:
I think we should prefer native ES features instead of Ramda helpers if they have the same interface, because:
- ES features are standardized and well-known, they can be learned once and used everywhere without third-party dependencies
- it should make bundle smaller because less code will be imported from Ramda package

Also `filter`, `map` with chaining in ES looks more readable:
```js
const modules = templateData
  .filter(R.propEq('kind', 'module'))
  .map(({ name }) => {
    return { name, out: `api/${name.replace(prefix, '')}` }
  })
```
than:
```js
const modules = R.map(({ name }) => {
  return { name, out: `api/${name.replace(prefix, '')}` }
}, R.filter(R.propEq('kind', 'module'), templateData))
```


nduchak:
I agree that this is cosmetic changes, maybe it's take us some perfomence and lower bundle size but it's not so critical. Also code looks inconsistent, so we need to decide do we need `ramda` at all in this case.

mpowaga:
I also prefer standard JS methods and chaining. On the other hand ramda is more powerful and has some handy functions not available in JS standard (like `R.pick` for example).

> it should make bundle smaller because less code will be imported from Ramda package 

@davidyuk How can we achieve this? 

1. custom ramda build
2. importing submodules (i.e. `import pick from 'ramda/pick'`)
3. tree shaking

AFAIK 2 option is not possible with ramda.

And as mentioned already it's not so critical at this point.

ricricucit:
yes, option 2 is not an option :)

davidyuk:
> by not actually getting rid of R completely

We shouldn't be getting rid of `R` completely because it has some features (like `R.pick`) that are not defined in ES, and using of this features reduces code duplication between projects, it the same way I'm using `lodash` in the Base app and other applications.

> therefore not solving a potentially good idea of entirely getting rid of a dependency.

but this PR removes dependencies on `ramda/map`, `ramda/reduce`, `ramda/filter` files.

@ricricucit thanks for article, I was not aware of the additional features that Ramda implements in `R.map`, are we using them somewhere?

davidyuk:
@mpowaga 
> 2. importing submodules (i.e. import pick from 'ramda/pick')

Currently, this is implemented by `babel-plugin-ramda` that SDK already uses.

> 3. tree shaking

As I know it will work (without `babel-plugin-ramda` plugin) if we will import Ramda by
```
import { toString, isNil, contains } from 'ramda/es'
```
instead of
```
import * as R from 'ramda'
```
because in the second case Webpack can't predict what Ramda features will be actually used (developer can write `R[methodName]` and so on).

ricricucit:
I'm fine with merging this. @nduchak @mpowaga ?

PR #122, opened 2018-10-29
Rename `CLI` to `Universal` and fix doc's

nduchak:


bin/aecontract.js
```diff
@@ -24,7 +24,7 @@
 
 'use strict'
 
-// We'll need the main client module `Ae` in the `Cli` flavor from the SDK.
+// We'll need the main client module `Ae` in the `Universal` flavor from the SDK.
 const { Cli: Ae } = require('@aeternity/aepp-sdk')
```
ricricucit:
```suggestion
const { Universal: Ae } = require('@aeternity/aepp-sdk')
```

bin/aewallet.js
```diff
@@ -22,7 +22,7 @@
 
 'use strict'
 
-// We'll need the main client module `Ae` in the `Cli` flavor from the SDK.
+// We'll need the main client module `Ae` in the `Universal` flavor from the SDK.
 const { Cli: Ae } = require('@aeternity/aepp-sdk')
```
ricricucit:
```suggestion
const { Universal: Ae } = require('@aeternity/aepp-sdk')
```

nduchak: Thank you.

PR #125, opened 2018-11-01
Aeproject compile

vsavovlime:


bin/aeproject/cli-commands/compile/compile.js
```diff
@@ -0,0 +1,57 @@
...
+
+async function compileAndPrint(file){
+    print('\r')
+    
+    try{
+        let client = await utils.getClient();
+        let code = readFile(file)
+        let contract = await client.contractCompile(code.toString());
+
+        print(`Contract '${file} has been succeffully compiled'`)
```
jsnewby: successfully

bin/aeproject/cli-commands/compile/compile.js
```diff
@@ -0,0 +1,57 @@
...
+ *  AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+ *  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+ *  LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+ *  OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ *  PERFORMANCE OF THIS SOFTWARE.
+ */
+require = require('esm')(module /*, options */) // use to handle es6 import/export
+import { printError, print } from '../../../utils/print'
+const utils = require('../../utils.js');
+const readFile = require('../../../utils/helpers').readFile;
```
maptuhec: I think this is fine for now, but we should keep one in mind for this and "print" if we split projects.

bin/aeproject/utils.js
```diff
@@ -31,7 +46,40 @@ const copyFileOrDir = (sourceFileOrDir, destinationFileOrDir, copyOptions = {})
   fs.copySync(sourceFileOrDir, destinationFileOrDir, copyOptions)
 }
 
+const getFiles = async function (directory, regex) {
```
maptuhec: Fix **test.js** file to use the function from the **utils** file instead the implementation in it

vsavovlime: Done

PR #127, opened 2018-11-02
Feature/remove cli from sdk

nduchak:


ricricucit:
@nduchak why is this including changes to `aeproject`?
@Perseverance @maptuhec should we maybe move `aeproject` out together with the `CLI`?

PR #143, opened 2018-11-14
Implement AENS native transaction

nduchak:
Implement AENS native transaction
Change bignumber libary


es/utils/bignumber.js
```diff
@@ -1,5 +1,5 @@
-import BigNumber from 'bignumber.js'
+const bignum = require('bignum')
```
ricricucit: let's use `import` for consistency

nduchak: You are right/ Forgot to change it

PR #145, opened 2018-11-20
Parse `fee` using bignum.js in nativeMode

nduchak:


ricricucit:
I don't understand why only the build on Jenkins is actually catching linting error........it's frustrating.

nduchak:
> I don't understand why only the build on Jenkins is actually catching linting error........it's frustrating.

My bad, forgot to run build before pushing.

ricricucit:
> My bad, forgot to run build before pushing.

oh. I did run it to double check, but I didn't realise I run it after your new push 😆

PR #154, opened 2018-11-27
bytes: Simplify `toBytes`

davidyuk:
`bitLength` is already calculated inside of `toArrayLike`: https://github.com/indutny/bn.js/blob/e69c617b3297b99aca429f30842e27979ef9beb5/lib/bn.js#L545


nduchak:
Not sure about that. We have the same implementation in python. Where we cover the array length.

ricricucit:
I'll wait for @nduchak to double check this. Looked fine to me...

nduchak:
We can merge it 👍

PR #155, opened 2018-11-28
Features/epoch 1.0.0 compatability

nduchak:


es/chain/epoch.js
```diff
@@ -26,7 +25,7 @@ async function sendTransaction (tx, options = {}) {
 }
 
 async function balance (address, { height, hash } = {}) {
-  return parseBigNumber((await this.api.getAccountByPubkey(address, { height, hash })).balance)
+  return (await this.api.getAccountByPubkey(address, { height, hash })).balance.toString()
```
davidyuk: Is `toString()` necessary if you are using `json-bigint` with `storeAsString` option?

nduchak: `json-bigint` convert to string only bignumber value. If balance not a `bignumber` it will not parse it.

PR #161, opened 2018-11-28
Use Buffer.toString('base64') instead of `base64-js` package

davidyuk:


nduchak:
@davidyuk Can we close that PR, it is includes in PR (https://github.com/aeternity/aepp-sdk-js/pull/162)?

davidyuk:
Firstly I had an idea to make them independent, but then I realized that decodeRaw accepts Uint array instead of Buffer.

PR #162, opened 2018-11-28
crypto: Use Buffer.equals and Buffer.toString('base64')

davidyuk:


davidyuk:
Firstly I had an idea to make them independent, but then I realized that `decodeRaw` accepts Uint array instead of Buffer.

PR #178, opened 2018-12-15
Fix network id propagation and Feed a Buffer to base58check encoding

ricricucit:


ricricucit:
Merging myself without review, because Saturday.

PR #183, opened 2018-12-17
Use `prepare` lifecycle event instead of `postinstall-build` package

### LINKED ISSUE #104: Drop deprecated `postinstall-build` package
> postinstall-build's behavior is now built into npm! You should migrate off of postinstall-build and use the new `prepare` lifecycle script with npm 5.0.0 or greater.

https://www.npmjs.com/package/postinstall-build

davidyuk:
Closes #104

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4397,12 +4397,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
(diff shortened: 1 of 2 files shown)


ricricucit:
Thank you @davidyuk :)

PR #187, opened 2018-12-20
Feature/pretty amount

nduchak:


es/chain/epoch.js
```diff
@@ -17,15 +17,17 @@
...
   const { waitMined } = R.merge(this.Chain.defaults, options)
   const { txHash } = await this.api.postTransaction({ tx })
   return waitMined ? this.poll(txHash, options) : txHash
 }
 
-async function balance (address, { height, hash } = {}) {
-  return (await this.api.getAccountByPubkey(address, { height, hash })).balance.toString()
+async function balance (address, { height, hash, format = true } = {}) {
+  const { balance } = await this.api.getAccountByPubkey(address, { height, hash })
+  return format ? formatBalance(balance) : balance.toString()
```
davidyuk: Looks like the balance should be shifted by 18 digits to right. Why `format` is true by default? I think balance firstly should be used in business logic and only then in the interface.

ricricucit: Thanks for double checking. 

We're gonna rewrite this anyway (as stated in the [CHANGELOG (v 1.2.1)](https://github.com/aeternity/aepp-sdk-js/blob/master/CHANGELOG.md))...but we tested this on the paperwallet app...and correctly served the purpose ATM.

I think the _actual_ solution will be returning strings like `14.6 ae` or `9000000000000000000 yanis` ...but for this we have to define the name of the units.

After this we'll probably say something like `client.balance(`PUBKEY`, {format: 'ae'})` with, as you propose, the smallest unit (called `yani`?) as default.

ricricucit: Also: this is an implementation that needs to be discussed with the product owner (@emil-apeunit @noandrea @jsnewby ?) ...and will need to be consistent across all SDKs

PR #188, opened 2018-12-20
Fix for Aepp stamp init

nduchak:


ricricucit:
@thepiwo can you do a PR for that?

PR #192, opened 2018-12-24
Simplify bigNumberToByteArray function

davidyuk:


nduchak:
@denis Some time ago you already do the same, but for some reason it's not work as expected. I was tested it using python-sdk and we have the different result.

davidyuk:
@nduchak Can you show a case in with my proposal works differently? probably this case should be added as a unit test

noandrea:
if ain't broken dont fixit

davidyuk:
You shouldn't invent the wheel, my proposal makes code easier to read and maintain.

PR #199, opened 2019-01-09
wallet example: Use network id provided by epoch node

davidyuk:


davidyuk:
Why gets closed? Is it important for examples to be compatible with previous versions of nodes?

davidyuk:
Anyway, this example won't get broken because default network id is [`ae_mainnet`](https://github.com/aeternity/aepp-sdk-js/blob/b1dce38d79a3d80e27bb66ecc8e4d3a22d2f2d52/es/account/index.js#L29)

nduchak:
Yes, it is. We support node which not provide networkId.

PR #201, opened 2019-01-10
Fix/contract stamp defaults

nduchak:


nikita-fuchs:
Is the `deposit` key still being used ?

nduchak:
Yes, we are using it in contract-create transaction.

nikita-fuchs:
Then we should also set this to 0 by default. It was the value that was deposited as the contract's initial balance, right?

nduchak:
Yes, you are right  👌

PR #207, opened 2019-01-11
Move high-level functions to wallet side

davidyuk:
required for https://github.com/aeternity/aepp-base/pull/303


davidyuk:
> Why do you need high-level function's executing on Wallet side?

To able determine aepp intention before calling of `sign` method. These are needed because currently, we can't parse raw transaction data.

nduchak:
If i understand correctly you need to handle `sign` method in `Aepp` side?
If yes you can override the sign method in Aepp stamp to not go through rpc

PR #208, opened 2019-01-11
Don't depend on Epoch explicitly

davidyuk:
To make possible initialise stamps with existing Epoch instance

required for https://github.com/aeternity/aepp-base/pull/303


davidyuk:
For me it is not necessary anymore, shall we close it?

PR #210, opened 2019-01-22
New Transaction builder

nduchak:
Implementation of new transaction builder 
Transaction serialization and deserialization going through SCHEMA.
Add new stamp `TransactionValidator`


es/tx/tx.js
```diff
@@ -37,8 +39,23 @@ async function spendTx ({ senderId, recipientId, amount, payload = '' }) {
 
   // Build transaction using sdk (if nativeMode) or build on `EPOCH` side
```
noandrea: Please change the references to `EPOCH` to `Aeternity Node`

PR #253, opened 2019-02-19
[WIP] Improve RPC to allow multiple channels and two-way communication

mpowaga:
1. Added `rpcChannel` parameter. This allows to use different protocols, i.e. `Wallet <- Aepp` communication is different than `Wallet <-> Channel`. Also should be possible now to create multiple instances of `Wallet`s.
2. `RpcClient` and `RpcServer` can be composed together to allow two-way communication.
3. Added `connectRpc` method so we have better control when the session is initialized (for example when iframe is loaded).


ricricucit:
@mpowaga can I close this?

PR #275, opened 2019-03-05
Improve channel rpc usage

mpowaga:
`balances` and `poi` methods can now be called at any point.


mpowaga:
@ricricucit That's a good idea. We need better integration of `Channel` with other stamps. How do you imagine the interface could look like?

PR #286, opened 2019-03-18
Identity provider PoC

nduchak:


nduchak:
Need to add new message `broadcastResponse`.

PR #303, opened 2019-04-01
Compiler improvements

nduchak:
Add ability to get account/balance on specific block hash/height.
Add test.
Add changeCompilerUrl method to Compiler stamp

#302

### DIFF
--- a/es/contract/compiler.js
+++ b/es/contract/compiler.js
@@ -48,7 +48,7 @@ async function contractGetACI (code, options = {}) {
   return this.http.post('/aci', { code, options }, options)
 }
 
-async function setCompilerUrl (url) {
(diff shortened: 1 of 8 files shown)


es/contract/compiler.js
```diff
@@ -48,6 +48,10 @@ async function contractGetACI (code, options = {}) {
   return this.http.post('/aci', { code, options }, options)
 }
 
+async function setCompilerUrl (url) {
```
mpowaga: @nduchak Does this function need to be declared as `async` even if it doesn't return anything?

nduchak: Yes, you are right.

nduchak: Fixed

Issue #320, opened 2019-04-10
State channels ping/pong


mpowaga:
Ping websocket connection every 10 seconds so it stays alive.

noandrea:
this is to keep alive the connection to the node for state channels

PR #324, opened 2019-04-10
feat(State Channels): Ping every 10 seconds to persist connection

mpowaga:
Connection can be manually closed with new "disconnect()" method

Implements #320


noandrea:
the error in the ci reports
```
No test report files were found. Configuration error?
Error when executing always post condition:
hudson.AbortException: No test report files were found. Configuration error?
```

PR #348, opened 2019-04-18
Add prototype of reverse iframe approach to example aepp

davidyuk:
The last commit of this PR contains a prototype of 'reverse-iframe' approach that we are going to implement in the desktop version of Base app and make it compatible with hybrid voting aepp.

This implementation doesn't require any changes in SDK, though would be better to integrate them into sdk somehow to make dapps building easier. 

ps: I was too lazy to rebase it on top of develop branch, so it contains changes from #347, I believe what that PR will get merged earlier than this one.

### DIFF
--- a/examples/connect-two-ae/aepp/src/components/Home.vue
+++ b/examples/connect-two-ae/aepp/src/components/Home.vue
@@ -147,6 +147,7 @@ import Aepp from 'AE_SDK_MODULES/ae/aepp'
 export default {
   data () {
     return {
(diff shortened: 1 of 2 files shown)


ricricucit:
@davidyuk with "reverse iFrame", you are talking about the possibility of getting info from the parent window? (AKA wallet)?

davidyuk:
@ricricucit No, the thing that you have mentioned can be called "usual iFrame". I am proposing to implement the possibility of getting info from the child window with the wallet opened in it.

PR #349, opened 2019-04-18
feat(ACI): Add `contract`, `address`, `record` types argument/result transformation

nduchak:
Allow pass contract address as sophia 'address' type


nduchak:
Implements: https://github.com/aeternity/aepp-sdk-js/issues/352

nduchak:
Implements: #352

PR #368, opened 2019-04-29
Init config for argon2 in browser

nduchak:


examples/connect-two-ae/identity/webpack.config.js
```diff
@@ -43,6 +44,7 @@ module.exports = {
   },
   devtool: process.env.NODE_ENV === 'prod' ? '' : 'eval-source-map',
   plugins: [
```
ricricucit: Not sure copying is the best idea. Can't this be solved by including the argon-2 in the JS rules [like this](https://github.com/aeternity/aepp-sdk-js/blob/develop/docs/usage.md#webpack-example) ?

nduchak: I am try to find something more elegant, but i don't. It's not so much info about ASM and WebAssembly.

nduchak: Fixed by using an `libsodium-wrapper` for argon2 in browser

PR #376, opened 2019-05-05
Clean up SDK dependencies

davidyuk:


davidyuk:
"serialize-javascript" is a dependency of "uglifyjs-webpack-plugin" package. Looks like changing pnpm to npm fixes build problem. Is this an issue of pnpm package manager or maybe it's cache on CI?

nduchak:
> "serialize-javascript" is a dependency of "uglifyjs-webpack-plugin" package. Looks like changing pnpm to npm fixes build problem. Is this an issue of pnpm package manager or maybe it's cache on CI?

I think it's CI issue.

PR #396, opened 2019-05-14
fix(ACI): Add ability to pass zero address as number.

nduchak:


es/contract/aci.js
```diff
@@ -188,7 +188,7 @@ function prepareSchema (type) {
     case SOPHIA_TYPES.string:
       return Joi.string().error(getJoiErrorMsg)
     case SOPHIA_TYPES.address:
-      return Joi.string().regex(/^(ak_|ct_|ok_|oq_)/).error(getJoiErrorMsg)
+      return Joi.alternatives([Joi.number().min(0).max(0), Joi.string().regex(/^(ak_|ct_|ok_|oq_)/).error(getJoiErrorMsg)])
```
mpowaga: This part doesn't make sense to me `Joi.number().min(0).max(0)` (probably because I don't know joi library). Maybe a comment would provide some hint?

nduchak: This rule check if the value is number and if it's value is 0. (typeof value === 'number' && value === 0)

PR #397, opened 2019-05-14
feat(Fortuna): Node 3.0.0 compatibility

nduchak:


es/node.js
```diff
@@ -119,7 +119,7 @@ const Node = stampit({
   }
 })
 
-const NODE_GE_VERSION = '1.4.0'
-const NODE_LT_VERSION = '3.0.0'
+const NODE_GE_VERSION = '2.5.0'
```
noandrea: we can keep it this less strict

PR #433, opened 2019-05-21
Feature/use contract interface for encoding call data

nduchak:


nduchak:
Close due to issue with compiler ACI generator

PR #434, opened 2019-05-21
feat(Node): Avoid usage of "universal-url" package

davidyuk:
Removing of `universal-url` package saves ~340kb.

#### Bundle report of Identity app example before:
Before
#### After:
After


nduchak:
@davidyuk Please fix test.

PR #441, opened 2019-05-24
feat(Compiler/ACI): Make ACI compatible with compiler 3.0.0

nduchak:
Add compatibility for both compiler versions (version >= 1.0.0 && version <= 4.0.0)


kraikov:
When `call(fn, args)` is invoked, nothing happens (the node isn't called) and `undefined` is returned immediately.

PR #446, opened 2019-05-29
tx: Avoid duplicate code

davidyuk:


es/tx/tx.js
```diff
@@ -30,313 +30,137 @@ import { buildTx, calculateFee } from './builder'
...
+    return ({
+      ownerId,
+      gasPrice,
+      ...otherFields,
+      ctVersion,
+      vmVersion: ctVersion.vmVersion,
+      abiVersion: ctVersion.abiVersion,
+      contractId: buildContractId(
+        ownerId,
+        (await this.prepareTxParams(TX_TYPE.oracleQuery, arguments[0])).nonce
```
nduchak: We calculate nonce twice for this tx???

davidyuk:
Extraction of common code makes easier to change it in the future because after that you need to change it only in one place.

After this refactoring, specific parts of some methods become more clear that makes some questions like "why oraclePostQueryTx have so unusual return value in comparison with other methods?" more obvious. It points to parts that probably should be refactored to get a better consistency.

I think it is better to have a relation between transaction type and epoch method/name of account field in a separate object than to have it spread in code. It has a more declarative style that is easier to read.

As for me, this code is still less complicated than in `utils/swagger.js`, both of them uses a similar approach.

In general, it is not required, but I think it would be good to have. Extraction of common code is an important part of programming.

es/tx/tx.js
davidyuk: yes it is true, I can fix it if this PR would be accepted in general

nduchak:
Close due to inactivity.

PR #448, opened 2019-05-29
fix(Ae): Fix exception when it used without Contract stamp

davidyuk:
> vue.runtime.esm.js:1886 TypeError: Cannot read property 'defaults' of undefined
    at Object._callee2$ (webpack-internal:///./node_modules/@aeternity/aepp-sdk/es/ae/index.js:129:71)


nduchak:
Can you describe where are you using `Ae` stamp

davidyuk:
> Can you describe where are you using `Ae` stamp

I was trying to investigate some issue in Base app be removing of connected stamps one by one. Initially, it was Wallet stamp from SDK.

PR #481, opened 2019-06-13
README: improve quick start

andrenarchy:
The current quick start doesn't give the user steps that he/she can just follow. In addition, the flavor concept further complicates it since this is very uncommon for a npm module – you usually just import the stuff need. Also, nothing was done with the imported `Ae` object and the list in step 3 contained links and you'd have to read a lot to get anything done.

I want to simplify this drastically. The first step you always need to do is to create a keypair. That's fairly easy but I couldn't find it in the docs so I adapted the README with this tiny bit of code. The user can then go ahead and top up the account. This gives the user an easy-to-follow process with a first success.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,20 +56,18 @@ npm i @aeternity/aepp-sdk@next
 npm i https://github.com/aeternity/aepp-sdk-js#develop
 ```
 
-### 2. Import (a chosen Flavor)
+### 2. Create a Keypair
 
-Import the right [flavor](docs/usage.md). For this example with get the `Universal` flavor, which contains all the features of the SDK:
-
-```js
-import Ae from '@aeternity/aepp-sdk/es/ae/universal' // or other flavor
+```javascript
+import { Crypto } from '@aeternity/aepp-sdk/es'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,20 +56,18 @@ npm i @aeternity/aepp-sdk@next
 npm i https://github.com/aeternity/aepp-sdk-js#develop
 ```
 
-### 2. Import (a chosen Flavor)
+### 2. Create a Keypair
```
ricricucit: I think this could work, but only as point nr. 3 (of 4), with 4 being "Give Yourself some _AEs_" and 2 keep being "Import a Flavor", because choosing and Importing a flavor is still a basic entry point for those that start using the SDK.

What do you think @andrenarchy ?

ricricucit: ...or maybe we can switch poin nr. 2 with point nr. 3? (account first?)

ricricucit:
closing due to inactivity and in favor of the [new PR 481](https://github.com/aeternity/aepp-sdk-js/pull/481)

PR #488, opened 2019-06-13
fix(TxBuilder): Fix serialization of payload

etharner:
In case of empty payload unpackTx leads to 'ba_Xfbg4g==' in payload. After that we're calling buildTx with that payload and in builded tx payload stored in exact this string instead of decoded into empty string.

### DIFF
--- a/es/tx/builder/index.js
+++ b/es/tx/builder/index.js
@@ -98,6 +98,7 @@ function serializeField (value, type, prefix) {
     case FIELD_TYPES.signatures:
       return value.map(Buffer.from)
     case FIELD_TYPES.payload:
+      return decode(value, 'ba')
     case FIELD_TYPES.string:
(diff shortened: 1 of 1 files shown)


nduchak:
The main idea that payload can be any type of data. And when we unpack transaction we don't know for sure what the data is that's why we encode it to base64check and prepend 'ba' prefix.
In you case you can simply decode the payload after unpacking the transaction.

davidyuk:
I think it has sense to make possible to call `buildTx` with arguments that `unpackTx` returns without need of any additional transformations as it was before `4.0.0`.

In the Base app, we using it to allow a user to set custom tx fee.

nduchak:
Fixed in PR: https://github.com/aeternity/aepp-sdk-js/pull/498

PR #511, opened 2019-06-25
swagger: Speedup initialisation

davidyuk:
I have checked how long does it takes to create an SDK instance in the Base app.

| | Before | After |
|---|---|---|
| On testnet | 4039ms | 1952ms |
| On mainnet | 3631ms | 1450ms |

So, this change makes SDK initialisation ~2 times faster. Also, it significantly reduces the amount of computations as I see from reports of Chrome profiler:

Screenshot of report taken before:
before
[full report](https://github.com/aeternity/aepp-sdk-js/files/3326646/before.txt)

After:
after
[full report](https://github.com/aeternity/aepp-sdk-js/files/3326647/after.txt)

These extra computations make Base app freezing when it started or if current network changed, especially on mobile devices. You can compare https://stage-identity.aepps.com/ and https://feature-speedup-sdk.origin.aepps.com/ by yourself to see the difference.


nikita-fuchs:
This is awesome, good work @davidyuk !

PR #512, opened 2019-07-01
Third-party Wallet Support

nduchak:


ricricucit:
@nduchak following, is what I see when I have browsers extensions active in chrome, together with some error that, ATM, [your PoC](https://github.com/nduchak/identity-provider-PoC) is throwing.

I think this is "noise" that pass through the `postMessage`, so we should probably try to have our own identifier to catch.

es/utils/aepp-wallet-communication/wallet-connection/browser-runtime.js
```diff
@@ -29,26 +29,21 @@ import AsyncInit from '../../async-init'
 import { getBrowserAPI } from '../helpers'
 
 function connect (onMessage, onDisconnect) {
-  if (this.port) throw new Error('You already connected')
-  this.port = getBrowserAPI().runtime.connect(this.connectionInfo.id)
+  if (this.port.onMessage.hasListeners()) throw new Error('You already connected')
```
shekhar-shubhendu: `You are`
But I'd say not to say `you` but say what it is i.e. `client` or `wallet` and also add what it is connected to.

es/utils/aepp-wallet-communication/wallet-detector.js
```diff
@@ -0,0 +1,75 @@
...
+ * @example import BrowserRuntimeConnection from '@aeternity/aepp-sdk/es/utils/wallet-connection/browser-runtime'
+ */
+import AsyncInit from '../async-init'
+import BrowserRuntimeConnection from './wallet-connection/browser-runtime'
+import { METHODS } from './schema'
+
+const wallets = {}
+
+const handleDetection = (onDetected) => (msg) => {
+  if (!msg || !msg.data || !msg.data.method || msg.data.type === 'webpackOk') return undefined
```
ricricucit: here we could probably have our own `msg.data.type` ...so we only check if `msg.data.type == 'ourDataType'`...?

es/utils/aepp-wallet-communication/wallet-detector.js
```diff
@@ -0,0 +1,82 @@
...
+  if (!method || !params) return
+  const ifExist = wallets.hasOwnProperty(params.id)
+  if (method === METHODS.wallet.readyToConnect && !ifExist) {
+    const w = {
+      ...params,
+      async getConnection () {
+        // if detect extension wallet or page wallet
+        return this.type === 'extension'
+          ? BrowserRuntimeConnection({ connectionInfo: this })
+          : BrowserWindowMessageConnection({ connectionInfo: this, origin: this.origin, target: isInIframe() ? window.parent : source })
```
nduchak: Fix origin assignment

PR #543, opened 2019-07-14
Add instructions for Build Tools

nikita-fuchs:
add note for windows users to prevent issues during sdk installation


shekhar-shubhendu:
@nikita-fuchs I think it should be this link instead https://www.npmjs.com/package/windows-build-tools

shekhar-shubhendu:
also instead of only for windows, we should have a general section asking users to install the build tools

nikita-fuchs:
Ah @shekhar-shubhendu great, I didn't know about those to be honest. What else are there for other OSs btw ? It's a long time ago for me, I remember for mac there was Xcode whatever, and, at least for ubuntu, build essentials or something? How would we structure all that, what do you guys think ?

shekhar-shubhendu:
@nikita-fuchs in general for Linux it will be `build-essentials` package. on ubuntu you can install via `apt`.
macos, you need to install xcode I think and then do `xcode-select --install`

nduchak:
Can i merge it and include in release?

nikita-fuchs:
wait a short sec, I'll add some proper information for the other OSs

nduchak:
Include this changes into release PR, due to Jenkins issue

PR #569, opened 2019-07-30
 Make Swagger a sync stamp

davidyuk:
Hello @davidyuk , is it necessary? And why you need that?


nduchak:
> Hello @davidyuk , is it necessary? And why you need that?

Ok now i get it. 👍

PR #611, opened 2019-08-18
install instead of add

eminogrande:
Should it not be install instead of add?


nikita-fuchs:
It's an alias for install: https://docs.npmjs.com/cli/install

PR #620, opened 2019-08-21
chore(example): Update sdk in example app

nduchak:


etharner:
I'm receiving this error when using example aepp and identity both updated to this branch
image

nduchak:
> I'm receiving this error when using example aepp and identity both updated to this branch
> image

You need to fund this account before deploying contract

davidyuk:
@nduchak but before it was not necessary, why you decided to change it?

nduchak:
> @nduchak but before it was not necessary, why you decided to change it?

What not necessary???Have tokens on account to make transaction?

davidyuk:
> > @nduchak but before it was not necessary, why you decided to change it?
> 
> What not necessary???Have tokens on account to make transaction?

I'm about this change: https://github.com/aeternity/aepp-sdk-js/commit/ccfbd6b3f53671a5b40a8e9a63e2d4cd75661a45#diff-19756c1304d22bf30c05d2b6624cc5faL52-R52

PR #645, opened 2019-09-03
example aepp: Add example of scanning a QR code using Base app

davidyuk:
https://github.com/aeternity/aepp-base/pull/1149


nduchak:
What about `identity` app which not have `readQRcode` fn?

davidyuk:
I see that it is a kind of proprietary extension to aepp-wallet communication implemented only in the Base app. I have made it to show how it works.

Maybe it shouldn't be in SDK repository, but we don't have enough time to maintain own version of the example app.

nduchak:
> I see that it is a kind of proprietary extension to aepp-wallet communication implemented only in the Base app. I have made it to show how it works.
> 
> Maybe it shouldn't be in SDK repository, but we don't have enough time to maintain own version of the example app.

Sorry, can't merge thing which will break our example

PR #672, opened 2019-09-18
Link broken

eminogrande:
https://github.com/aeternity/aepp-sdk-js/blob/develop/examples/browser/connect-two-ae needs to be fixed but i don't know the right link.


shekhar-shubhendu:
@eminogrande you can use this link `browser/vuejs/connect-two-ae/README.md`

PR #800, opened 2019-11-20
tx: Sync nonce and increment internal nonce for every transaction

### LINKED ISSUE #291: Implement local node management
Issue tracker is **ONLY** used for reporting bugs. NO NEW FEATURE ACCEPTED! Use [stackoverflow](https://stackoverflow.com/tags/aeternity) for supporting issues.

## Steps to Reproduce

1.
2.
3.
4.

davidyuk:
supposed to fix #291

### DIFF
--- a/es/tx/tx.js
+++ b/es/tx/tx.js
@@ -410,9 +410,14 @@ async function calculateTtl (ttl = TX_TTL, relative = true) {
  * @return {number} Next Nonce
  */
 async function getAccountNonce (accountId, nonce) {
-  if (nonce) return nonce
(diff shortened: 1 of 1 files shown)


davidyuk:
Looks like the node rejects transactions with the nonce bigger than the last one + 1. I have some idea of how it can be fixed, I'll implement it later.

noandrea:
looks like that the amount of nonce increments are a configurable value for the node, and not limited to +1

davidyuk:
But how are nodes getting consensus if different nodes can require to increment nonces by different values?

Issue #851, opened 2020-01-08
Add functionality for waiting for transaction confirms


nduchak:
**Describe the solution you'd like**
Ability to wait for `N` blocks after transaction accepting

PR #874, opened 2020-01-22
feat(Chain): add new method `waitFOrTxConfirm`

nduchak:
Add new option `{ confirm: 3 }` to all of `high lvl SDK API`. Add tests. Adjust docs


nduchak:
Resolve #851

Issue #883, opened 2020-01-29
Add signature delegation helpers


nduchak:
**Is your feature request related to a problem? Please describe.**
Provide an easy way to prepare a signature to delegate actions in Sophia SC

PR #895, opened 2020-02-05
Feat(Contract): helper for Oracle and AENS signature delegation

nduchak:


nduchak:
Resolve #883

es/utils/string.js
```diff
@@ -56,6 +56,16 @@ export function isHex (str) {
   return !!(str.length % 2 === 0 && str.match(/^[0-9a-f]+$/i))
 }
 
+/**
+ * Check whether a string is Aeternity address
+ * @param {string} str String to validate.
+ * @return {boolean} True if the string is valid AE address
+ */
+export function isAeAddress (str) {
```
mradkov: This only checks if the string is prefixed with 2 char string before the `_`. E.g. `ak_`

Shouldn't it also check for `Crypto.isAddressValid` ?

nduchak: Yes, you are right

Issue #902, opened 2020-02-17
Wallet ID is missed in Identity example in Chrome


davidyuk:
https://github.com/aeternity/aepp-sdk-js/blob/c5fa163a6411e09f304702ae3b9e31caf17b93de/es/utils/aepp-wallet-communication/rpc/wallet-rpc.js#L276

`chrome.runtime` is defined in Chrome's global scope of a usual tab (not an extension), so the condition above will be truly, but `chrome.runtime.id` is not defined, it will lead to sending the wallet info without id that breaks the aepp side.

nduchak:
Actually it's doesn't break AEPP, because when you are using Iframe/reverse iframe you don't need this id, this is required only for extension connection

Issue #899, opened 2020-02-17
example/identity depends on the not available version of SDK


davidyuk:
https://github.com/aeternity/aepp-sdk-js/blob/c5fa163a6411e09f304702ae3b9e31caf17b93de/examples/browser/vuejs/connect-two-ae/identity/package.json#L16

nduchak:
Thanks for opening. Will fix it soon

davidyuk:
Also, I can't build identity example after updating of sdk:

<details>
  <summary>Console log</summary>
  
```bash
$ git reset HEAD --hard
HEAD is now at b70c79fb Merge pull request #888 from aeternity/release/7.0.0
$ npm i @aeternity/aepp-sdk@7.0.0

> fsevents@1.2.11 install /Users/denis/Documents/aepp-sdk-js/examples/browser/vuejs/connect-two-ae/identity/node_modules/fsevents
> node-gyp rebuild

  SOLINK_MODULE(target) Release/.node
  CXX(target) Release/obj.target/fse/fsevents.o
  SOLINK_MODULE(target) Release/fse.node

> websocket@1.0.31 install /Users/denis/Documents/aepp-sdk-js/examples/browser/vuejs/connect-two-ae/identity/node_modules/websocket
> (node-gyp rebuild 2> builderror.log) || (exit 0)

  CXX(target) Release/obj.target/bufferutil/src/bufferutil.o
  SOLINK_MODULE(target) Release/bufferutil.node
  CXX(target) Release/obj.target/validation/src/validation.o
  SOLINK_MODULE(target) Release/validation.node

> core-js@2.6.11 postinstall /Users/denis/Documents/aepp-sdk-js/examples/browser/vuejs/connect-two-ae/identity/node_modules/babel-runtime/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> core-js@3.6.4 postinstall /Users/denis/Documents/aepp-sdk-js/examples/browser/vuejs/connect-two-ae/identity/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> ejs@2.7.4 postinstall /Users/denis/Documents/aepp-sdk-js/examples/browser/vuejs/connect-two-ae/identity/node_modules/ejs
> node ./postinstall.js

Thank you for installing EJS: built with the Jake JavaScript build tool (https://jakejs.com/)

+ @aeternity/aepp-sdk@7.0.0
[...]

PR #903, opened 2020-02-17
Feat(Signing): Message Signing

nduchak:


nduchak:
Resolve https://github.com/aeternity/aepp-sdk-js/issues/904

nduchak:
Resolve #902, #899,  aeternity/tipwallet#87

es/utils/crypto.js
```diff
@@ -359,18 +359,18 @@ export function prepareTx (signature, data) {
 }
 
 export function personalMessageToBinary (message) {
-  const p = Buffer.from('æternity Signed Message:\n', 'utf8')
+  const p = Buffer.from('aeternity Signed Message:\n', 'utf8')
```
davidyuk: what is the problem with æ? As I know, we are using it on Ledger

mradkov: Hans suggested we use “ae” instead as otherwise we can face problems in the future.

PR #911, opened 2020-02-21
feat(AEX-2): Use `hex` string for `signMessage` response

nduchak:
Extend `verifyMessage` with ability to accept `buffer` or `hex`


nduchak:
Resolve https://github.com/aeternity/tipwallet/issues/90#issuecomment-589628127

PR #921, opened 2020-02-26
Refactor rpc-related stuff

davidyuk:


davidyuk:
@nduchak I have something to propose on top of that. Did you manage to check it?

PR #933, opened 2020-03-03
[WIP]feat(Tx): TX object wrapper

nduchak:


nduchak:
Closes: #933

nduchak:
> I guess it will be more helpful if we can also sign arbitrary data passed as hex instead of only Buffer? It will be easier to integrate and less error prone for some apps.

Resolved

PR #961, opened 2020-04-10
Refactor/move address helper to crypto

nduchak:


nduchak:
> Is the `getAddressFromPriv` being moved or copied to `Crypto` stamp?
> 
> Is this preserving backward compatibility?

It was copied, so it's backward compatible
Will remove duplicate in 8.0.0 version

PR #967, opened 2020-04-22
feat(Node): Add `debug` option to Channel stamp

nduchak:


.env
```diff
@@ -1,2 +1,2 @@
-NODE_TAG=v5.5.0
+NODE_TAG=v5.5.2
```
mradkov, line 1: `5.5.3`?

Issue #962, opened 2020-04-15
Call contract by AENS name


mradkov:
Currently its not possible to call a contract by an AENS name pointed to it.

https://github.com/aeternity/aepp-sdk-js/blob/develop/es/contract/aci/index.js#L114

As a workaround we need to first resolve the pointer and then apply the contract address.

It will be much better UX if this can happen in the SDK so a call to a contract pointed by name can happen like so:

```javascript
const myContract = await sdk.getContractInstance(CONTRACT_SOURCE, {  contractAddress: 'example.chain' });
```

Or better of - have `aensName` as optional param:

```javascript
const myContract = await sdk.getContractInstance(CONTRACT_SOURCE, {  aensName: 'example.chain' });
```

nduchak:
Resolved: #1015

PR #1015, opened 2020-06-01
Feat(Contract/ACI): Call contract using AENS name

nduchak:
- Resolve contract address from AENS name using Contract/ACI API
- Simplify usage of Wallet detector(make `connection` argument optional)


nduchak:
Closes: #962

PR #1018, opened 2020-06-02
fix(AEX-2): Handler always as Promise

nduchak:


etharner:
With this fixes it leads now to this error:
image
As I suggested, returning of promise will work for last `getHandler` in `handleMessage`

PR #1025, opened 2020-06-12
feat(Account): Build signature from transaction hash

nduchak:


nduchak:
Closes https://github.com/aeternity/aepp-sdk-js/issues/1026

PR #1031, opened 2020-06-16
Simplify Travis and docker-compose

davidyuk:


.travis.yml
```diff
@@ -1,71 +1,30 @@
-branches:
-  only:
-    - master
-    - develop
```
davidyuk, line 4: Can we disable "Build pushed pull requests" in Travis configs instead?
Screenshot 2020-06-17 at 10 58 20

PR #1034, opened 2020-06-17
Remove OracleNodeAPI wrapper

davidyuk:
OracleNodeApi is a wrapper of node interface (`sdk.api.*`) without any additional logic, I don't think that we need it. To use node api directly seems clearer for me because I know that there are no additional overheads, and I'm just making a request to the node. Though this is a breaking change, so probably better to postpone it until the next major release.


nduchak:
The old RPC stamps depend on this wrapper

davidyuk:
Breaking changes:
- removed OracleNodeAPI

Migration guide:
- use `api.getOracleByPubkey` instead of `getOracle`
- use `api.getOracleQueriesByPubkey` instead of `getOracleQueries`
- use `api.getOracleQueryByPubkeyAndQueryId` instead of `getOracleQuery`

thepiwo:
the logs of changes should probably be also in the changelog as described here: https://github.com/aeternity/aepp-sdk-js/pull/1034#issuecomment-653388894

mradkov (approved):
Changes look good.

PR #1038, opened 2020-06-19
tests: Avoid debug output

davidyuk:


davidyuk:
> It would be nice if we can have this output optional via some flag.

It is really easy to write such stuff on demand, I'm doing it often locally, but this shouldn't be pushed to the repository, even with a debug flag. I assume that some of this `console.log`s were pushed accidentally.

PR #1039, opened 2020-06-20
Refactor tests

davidyuk:
https://stackoverflow.com/questions/23492043/change-default-timeout-for-mocha


nduchak:
account2 is used in tests, try to find “ onAccount” in test folder

davidyuk:
`onAccount` mentioned 82 times in tests folder, can you point a specific use case?
I have assumed that `account2` is not used because it is not imported anywhere, also tests did not fall after removing it 🤷‍♀️
I would suggest defining `account2` exactly in tests that are using it.

PR #1041, opened 2020-06-21
Drop old aepp-wallet RPC interface

davidyuk:
Breaking changes:
- old aepp-wallet RPC is completely removed

Should we also rename `RpcAepp` and `RpcWallet` to `Aepp` and `Wallet` respectively?


davidyuk:
The current version of Base app depends on the old interface, but I think we can fix it by the next version of sdk. Do we need the support of old interface in the base aepp after updating to sdk@8? It would be tricky after removing the old rpc support from sdk. So, I propose to merge this PR only if we don't need the old rpc support in future versions of Base aepp.

davidyuk:
@mradkov can we merge this one?

mradkov (approved):

PR #1042, opened 2020-06-22
Combine Accounts and Selector into AccountMultiple

davidyuk:
Breaking changes:
- `Account` renamed to `AccountBase`
- `Accounts` and `Selector` combined into `AccountMultiple`
- `Selector.address` renamed to `selectedAddress`


davidyuk:
The `Selector` and `Accounts` stamps depend on each other ([selector](https://github.com/aeternity/aepp-sdk-js/blob/2ae757cd4c5701f3fedf3f0b5280d5862a582c19/es/account/selector.js#L97), [accounts](https://github.com/aeternity/aepp-sdk-js/blob/ef895eea33e4de9ab3190317a3be4e9444fa8112/es/accounts.js#L67)) and not used separately in sdk, so I propose to combine them, it makes the code clearer.

Also, I proposing to avoid a practice to hold an interface stamp in `index` file and name it `base` instead (`interface` can't be used because it is not always just an interface).

mradkov (approved):

PR #1045, opened 2020-06-29
Add lint script

davidyuk:


es/utils/aepp-wallet-communication/connection/browser-runtime.js
```diff
@@ -55,10 +55,10 @@ function disconnect () {
  */
 function connect (onMessage, onDisconnect) {
   if (this.isConnected()) throw new Error('You already connected')
-  this.handler = ((msg, source) => {
+  this.handler = (msg, source) => {
     if (this.debug) console.log('Receive message: ', msg)
     onMessage(msg, source)
-  }).bind(this)
+  }
```
thepiwo, line 61: sure this is not breaking?

examples/browser/extension/utils/build.js
```diff
@@ -1,9 +1,9 @@
-var webpack = require("webpack"),
-    config = require("../webpack.config");
+var webpack = require('webpack')
+var config = require('../webpack.config')
```
thepiwo, line 2: why `var` over `const`? maybe the linter should notice?

examples/browser/extension/utils/webserver.js
```diff
@@ -1,35 +1,35 @@
...
+var WebpackDevServer = require('webpack-dev-server')
+var webpack = require('webpack')
+var config = require('../webpack.config')
+var env = require('./env')
+var path = require('path')
 
-var options = (config.chromeExtensionBoilerplate || {});
-var excludeEntriesToHotReload = (options.notHotReload || []);
+var options = (config.chromeExtensionBoilerplate || {})
+var excludeEntriesToHotReload = (options.notHotReload || [])
```
thepiwo, line 8: why `var` over `const`? maybe the linter should notice?

es/utils/aepp-wallet-communication/connection/browser-runtime.js
davidyuk: Yes, https://stackoverflow.com/a/33308151/6176994
Also, I have checked it in the browser:
```js
> (() => this).bind({ test: true })()
Window {parent: Window, opener: null, top: Window, length: 0, frames: Window, …}
> (function() { return this }).bind({ test: true })()
{ test: true }
```
It looks like a tricky question at the interview 😅

examples/browser/extension/utils/build.js
davidyuk: Probably the linter assumes that it is ES5 code. I would like to revisit examples later.

PR #1047, opened 2020-07-01
Refactor state channels

davidyuk:


davidyuk:
@mradkov @thepiwo I think this one is ready for review

es/channel/internal.js
```diff
@@ -27,31 +26,24 @@ const PING_TIMEOUT_MS = 10000
...
 const pongTimeoutId = new WeakMap()
-const fsmId = new WeakMap()
+export const fsmId = new WeakMap()
 
-function channelURL (url, params) {
-  const paramString = R.join('&', R.values(R.mapObjIndexed((value, key) =>
-    `${pascalToSnake(key)}=${encodeURIComponent(value)}`, params)))
-
-  return `${url}?${paramString}`
-}
```
mradkov, line 52: Isn't this used anywhere?

davidyuk: It is used only in this file, in `initialize` function, otherwise, it will be rejected by linter.

davidyuk:
> Do we have any example where we can try this?

I have tried this branch in the cont-toss-game (https://github.com/aeternity/coin-toss-game/tree/superhero-demo, https://github.com/aeternity/ae-channel-service/pull/113), it works fine with these changes. 

> Does the test coverage also cover this?

Yes, it is heavily covered in this file: https://github.com/aeternity/aepp-sdk-js/blob/9494365840f47dbad40c515f1e7b6b4f91928ce1/test/integration/channel.js

PR #1048, opened 2020-07-01
Refactor contract, oracle, chain

davidyuk:


es/ae/oracle.js
```diff
@@ -61,27 +62,23 @@ async function getOracleObject (oracleId) {
...
-    if (newQueries.length) {
-      quriesIds = [...quriesIds, ...newQueries.map(a => a.id)]
-      onQuery(newQueries)
-    }
+function pollForQueries (oracleId, onQuery, { interval = 5000 } = {}) {
+  const sentQueryIds = new Set()
+  const fetchQueries = async () => {
+    const queries = ((await this.getOracleQueries(oracleId)).oracleQueries || [])
+      .filter(({ id }) => !sentQueryIds.has(id))
+    queries.forEach(({ id }) => sentQueryIds.add(id))
```
thepiwo, line 76: I'd refactor this to use map again.

(not tested)
```javascript
const allQueries = (await this.getOracleQueries(oracleId)).oracleQueries || [])
const queries = [...new Set(allQueries.map(({ id }) => id))]
```

es/ae/oracle.js
```diff
@@ -61,27 +62,23 @@ async function getOracleObject (oracleId) {
...
-  async function pollQueries () {
-    const { oracleQueries: q } = await this.getOracleQueries(oracleId)
-    const newQueries = q.filter(({ id }) => !quriesIds.includes(id))
-    if (newQueries.length) {
-      quriesIds = [...quriesIds, ...newQueries.map(a => a.id)]
-      onQuery(newQueries)
-    }
+function pollForQueries (oracleId, onQuery, { interval = 5000 } = {}) {
+  const sentQueryIds = new Set()
+  const fetchQueries = async () => {
```
thepiwo, line 73: fetch implies to me something is returned, but we just want the `onQuery` to be triggered

es/ae/oracle.js
```diff
@@ -121,22 +118,15 @@ async function getQueryObject (oracleId, queryId) {
...
-  }
-  async function probe (left) {
-    const query = await this.getOracleQuery(oracleId, queryId)
-    if (query.response !== emptyResponse) {
-      return { response: query.response, decode: () => decodeBase64Check(query.response.slice(3)) }
-    }
-    if (left > 0) {
-      await pause(interval)
-      return probe.bind(this)(left - 1)
+  for (let i = 0; i < attempts; i++) {
```
thepiwo, line 121: are you sure the for correctly awaits the promise here for each iteration?

I'd maybe go for a more high-level approach.

1. use a range array instead of `for i` (e.g. as in https://github.com/aeternity/aepp-governance/blob/master/governance-server/src/util.js#L45)
2. use an async reduce to iterate and eventually break the iteration (maybe something similar to https://github.com/aeternity/aepp-governance/blob/master/governance-server/src/util.js#L58 can be helpful)

thepiwo: This could also be combined in a util function, e.g. as `attemptWithPause(() => ...)` passing the function, where e.g. a rejected promise or returned false is the break condition

es/chain/node.js
```diff
@@ -115,54 +108,35 @@ async function height () {
...
-      return current
-    }
-    if (left > 0) {
-      await pause(interval)
-      return probe(left - 1)
-    }
-    throw Error(`Giving up after ${attempts * interval}ms, current=${current}, h=${h}`)
+async function awaitHeight (height, { interval = 5000, attempts = 20 } = {}) {
+  let currentHeight
+  for (let i = 0; i < attempts; i++) {
```
thepiwo, line 113: same as above applies

thepiwo:
I didn't notice the minor breaking changes, maybe a summary of what the PR does with a list of breaking changes would be helpful in the PR description.

es/ae/oracle.js
davidyuk: I'm proposing a better naming in 5164358a931215ef629608c6d1369f302d966c75

es/ae/oracle.js
davidyuk: It looks like we don't need to store all previous queries, just the last one instead, and pass it as `from` parameter to GetOracleQueriesByPubkey. I would like to check it later, proposing to leave as it is now.
https://github.com/aeternity/aepp-sdk-js/issues/1055

es/ae/oracle.js
davidyuk: > are you sure the for correctly awaits the promise here for each iteration?

Yes, I think so. Actually, I'm rarely writing a `for` cycle, but I think it is a good place to use it. I prefer this because it uses just plain js.

davidyuk:
I have rebased this PR on develop and fixed tests.

> I didn't notice the minor breaking changes

The only one was a making pollForQueries to run in sync, I have made it return a dumb promise to be able to merge it earlier.

davidyuk:
I have just restarted tests and they have passed 🤷‍♀️

PR #1052, opened 2020-07-06
Build docs automatically using Travis

davidyuk:
- avoid inconsistency between docs and code (no need to rebuild docs manually for each PR/release)
- avoid duplicate results in project-wide search
- will check for correct docs definition using travis (to handle issues like af8d1c950116cb1ca01fff7a923f5f0a4b18783a before merging)

Need to add http://aeternity.github.io/aepp-sdk-js/ to the root readme and instead https://dev.aepps.com in project description.


mradkov:
> Need to add http://aeternity.github.io/aepp-sdk-js/ ...

This actually does not work. 🤔

davidyuk:
Here it is deployed on my fork: http://davidyuk.github.io/aepp-sdk-js/
Can we unbind aeternity.com from GitHub pages of this repo? Probably we will have issues because of it.

mradkov:
> Here it is deployed on my fork: http://davidyuk.github.io/aepp-sdk-js/
> Can we unbind aeternity.com from GitHub pages of this repo? Probably we will have issues because of it.

@venimus can you take a look here and suggest change in the domain config or sth else?

davidyuk:
@mradkov so, what is the outcome of gh-pages stuff? Should we postpone merging this in the 7.7.0 milestone?

mradkov:
> @mradkov so, what is the outcome of gh-pages stuff? Should we postpone merging this in the 7.7.0 milestone?

Yeah, maybe we can omit this for now. Waiting for @venimus or @dincho to check the domain config.

davidyuk:
Is the master branch deployed to https://aeternity.com/aepp-sdk-js/? I don't have permission to check that but seems that it is not a `gh-pages` branch.

I don't have an opinion on what domain the docs should be published, but for me, the default handler of gh-pages works more obvious than VuePress, can I choose the default handler somehow?

dincho:
@mradkov is admin of this repo btw

@davidyuk currently the GH pages of this repo is turned off

As @venimus mentioned there are two options:
- org-domain.tld/repo-name, in our case aeternity.com/aepp-sdk-js
- custom domain, e.g. ae-js-sdk.com (yeah ugly but just and example), so the most practcal a subdomain, like the erlang node does docs.aeternity.io (it's not on GH pages, but you get the point)

mradkov (approved):

PR #1056, opened 2020-07-09
example: Add disconnect button on wallet side

davidyuk:
@CedrikNikita have I missed something?


CedrikNikita:
Yes, it's important to have this in the examples.

PR #1057, opened 2020-07-13
Add support of bits type

davidyuk:
it was requested by @dincho


mradkov:
Is this still a draft?

davidyuk:
yes, I need to think about how correctly represent this type

mradkov:
The calldata lib already supports `Bits` type so we can close this PR.

PR #1058, opened 2020-07-15
index: Export in tree shaking friendly way

davidyuk:
Currently, to import something from `crypto` module you have to import it completely, like:
```js
import { Crypto } from '@aeternity/aepp-sdk-js';
Crypto.isAddressValid(...)
```
Tree shaking works better if imported only a specific thing, I'm proposing to make it possible by inlining `crypto` exports into `index` exports, making user import it by:
```js
import { isAddressValid } from '@aeternity/aepp-sdk-js';
isAddressValid(...)
```

### DIFF
--- a/docs/guides/aens.md
+++ b/docs/guides/aens.md
@@ -121,7 +121,7 @@ Note:
 In case there is an auction running for a name you want to claim you need to place a bid.
 
 ```js
-import { TxBuilderHelper } from '@aeternity/aepp-sdk'
+import { computeBidFee, computeAuctionEndBlock } from '@aeternity/aepp-sdk'
 
 const name = 'auctiontest1.chain'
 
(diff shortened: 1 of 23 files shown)


mradkov:
> Currently, to import something from `crypto` module you have to import it completely, like:
> 
> ```js
> import { Crypto } from '@aeternity/aepp-sdk-js';
> Crypto.isAddressValid(...)
> ```
> 
> Tree shaking works better if imported only a specific thing, I'm proposing to make it possible by inlining `crypto` exports into `index` exports, making user import it by:
> 
> ```js
> import { isAddressValid } from '@aeternity/aepp-sdk-js';
> isAddressValid(...)
> ```

I like the idea, but we should also consider that the `Crypto.something` notation is kind-of more understandable that you are dealing with the actuall crypto stuff - signatures, keypairs etc.

davidyuk:
We can just improve naming somehow, making obvious that the functions are crypto-related.

thepiwo (commented):
cool changes, whats the state of this?

davidyuk:
Several devs checked this and didn't find any major issues, so I think we can proceed

PR #1059, opened 2020-07-16
Drop old names support, split ensureNameValid and isNameValid

davidyuk:
Breaking changes:
- removed `nameHash` from `builder/helpers`
- `isNameValid` won't throw exceptions
- removed `AENS_NAME_DOMAINS` from `builder/schema`
- removed default export of `builder/helpers`
Features:
- added `ensureNameValid` function that will throw exception if name is not valid


mradkov (approved):
LGTM. We need to make sure we notify properly the users about the breaking changes.

PR #1060, opened 2020-07-16
transferFunds: Accept onAccount option

### LINKED ISSUE #1046: transferFunds: Implement support of onAccount option
https://github.com/aeternity/superhero-wallet/pull/400#discussion_r446865923

davidyuk:
closes #1046

### DIFF
--- a/es/ae/index.js
+++ b/es/ae/index.js
@@ -39,7 +39,7 @@ import { AE_AMOUNT_FORMATS } from '../utils/amount-formatter'
  * @param {String} tx - Transaction
  * @param {Object} [options={}] options - Options
  * @param {Object} [options.verify] verify - Verify transaction before broadcast, throw error if not valid
- * @return {String|String} Transaction or transaction hash
+ * @return {Object} Transaction
  */
 async function send (tx, options = {}) {
   const opt = R.merge(this.Ae.defaults, options)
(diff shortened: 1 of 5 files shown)


es/ae/index.js
```diff
@@ -59,50 +59,51 @@ async function signUsingGA (tx, options = {}) {
...
-  return this.send(spendTx, opt)
+  const opt = { ...this.Ae.defaults, ...options }
+  const recipientId = await this.resolveName(recipientIdOrName, 'ak', opt)
+  const senderId = await this.address(opt)
+  const balance = new BigNumber(await this.balance(senderId))
+  const desiredAmount = balance.times(fraction)
+  const { tx: { fee } } = TxBuilder.unpackTx(
+    await this.spendTx({ ...opt, senderId, recipientId, amount: desiredAmount })
+  )
+  // TODO: Reducing of the amount may reduce transaction fee, so this is not completely accurate
```
mradkov, line 104: Why is this the case?

test/integration/accounts.js
```diff
@@ -64,12 +65,41 @@ describe('Accounts', function () {
...
+      balanceBefore.should.be.not.eql(balanceAfter)
+      amount.isZero().should.be.equal(true)
+    })
+
+    it('spends 25% of balance', async () => {
+      const { balanceBefore, balanceAfter, amount, fee } = await spend(0.25)
+      amount.times(4).should.be.eql(balanceBefore)
+      balanceAfter.plus(amount).plus(fee).should.be.eql(balanceBefore)
+    })
+
```
mradkov, line 92: Can you also add a test for spending 50% of the balance? Not sure why it was removed.

es/ae/index.js
davidyuk: If we use `balance.minus(fee)` instead of the amount that we have calculated fee for then we may pay an extra fee because the fee depends on the length of the amount field. This should be a rare case, and the extra fee should be insignificant, so I don't think that is worth considering this case. Maybe we need to remove the `TODO` label 🙂

test/integration/accounts.js
davidyuk: In general, I see 4 classes of tests of the percentage option:
- value out of range (no covered currently)
- minimum value
- maximum value
- value in the range from 0 to 100%

25% and 50% are both from the last class. I have chosen 25 because it looks a bit more interesting than 50%, but now I think would be better to choose a completely random one, like 68.97% 😄 Also, these tests are quite expensive so I don't see reasons to include several values from the same class of tests. How would like to proceed with that?

mradkov: > In general, I see 4 classes of tests of the percentage option:
> 
> - value out of range (no covered currently)
> 
> - minimum value
> 
> - maximum value
> 
> - value in the range from 0 to 100%
> 
> 
> 
> 25% and 50% are both from the last class. I have chosen 25 because it looks a bit more interesting than 50%, but now I think would be better to choose a completely random one, like 68.97% 😄 Also, these tests are quite expensive so I don't see reasons to include several values from the same class of tests. How would like to proceed with that?

68.97% looks nice 🤔✌️

davidyuk: This is a good test, applying it I have found a rounding bug.

PR #1066, opened 2020-08-16
bigNumberToByteArray: Avoid unexpected behaviour by throwing exception

davidyuk:
Implementing #1060 I have found that passing a non-integer amount value looks working, but actually the amount value is gets broken instead of being rounded somehow. `BigNumber.toString(16)` returns a hex value separated with a dot for non-integer values that get passed to `Buffer.from` that just ignores the part after `.`:
```
> Buffer.from('390cf7311ac9f5ab9746ab.2ab9f559b3d07c84b5dd', 'hex')
<Buffer 39 0c f7 31 1a c9 f5 ab 97 46 ab>
```
Alignment by adding a leading 0 also may not work properly in this case.

`bigNumberToByteArray` is handling non-integers not properly, so I propose to throw an exception then.

### DIFF
--- a/es/utils/bytes.js
+++ b/es/utils/bytes.js
@@ -63,6 +63,7 @@ export function rightPad (length, inputBuffer) {
  * @return Buffer
  */
 export function bigNumberToByteArray (x) {
+  if (!x.isInteger()) throw new Error(`Unexpected not integer value: ${x.toFixed()}`)
   let hexString = x.toString(16)
   if (hexString.length % 2 > 0) hexString = '0' + hexString
   return Buffer.from(hexString, 'hex')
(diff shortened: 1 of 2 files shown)


davidyuk:
It looks like sdk uses `bigNumberToByteArray` properly in all cases, but the end-user may ask sdk to transfer for example 123.55 aettos and before this fix the sdk would send 7 aettos (`BN(123.55)` => `7b.8ccccccccccccccccccd` => `Buffer.from('07b.8cc...', 'hex')` => `[07]`) and with this fix sdk will reject that attempt. I assume that the user will figure out that he needs to round the token value somehow after seeing the exception.

PR #1073, opened 2020-09-27
Height query optimisation

gorbak25:
Inspired by the governance aepp
image
PR on top of #1072

### DIFF
--- a/es/chain/node.js
+++ b/es/chain/node.js
@@ -105,7 +105,14 @@ async function tx (hash, info = true) {
 }
 
 async function height () {
-  return (await this.api.getCurrentKeyBlockHeight()).height
+  try {
+    if (this.height_query_promise === undefined) {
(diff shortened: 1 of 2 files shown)


mradkov (approved):

es/chain/node.js
```diff
@@ -105,7 +105,14 @@ async function tx (hash, info = true) {
 }
 
 async function height () {
-  return (await this.api.getCurrentKeyBlockHeight()).height
+  try {
+    if (this.height_query_promise === undefined) {
```
davidyuk, line 109: this naming looks super unusual for js, shouldn't it be something like `_heightPromise`?

davidyuk (commented):
This promise should be updated if the height is queried on a different network, otherwise, this may bring some bugs to end-users.

I would consider implementing a general caching layer, I don't think that we have this issue only with this endpoint.

PR #1079, opened 2020-10-20
Fixed error throw on assertedType()

filipdujmusic:
Was this a bug? Please review

### DIFF
--- a/es/utils/crypto.js
+++ b/es/utils/crypto.js
@@ -444,7 +444,7 @@ export function assertedType (data, type, forceError = false) {
   if (RegExp(`^${type}_.+$`).test(data)) {
     return data.split('_')[1]
   } else {
-    if (!forceError) throw Error(`Data doesn't match expected type ${type}`)
+    if (forceError) throw Error(`Data doesn't match expected type ${type}`)
     return false
   }
 }


davidyuk (commented):
I would rather rename it to `omitError`. I can't get that `force` means in this context.

filipdujmusic:
> I would rather rename it to `omitError`. I can't get that `force` means in this context.

Naming should be more clear yes, because it will affect other parts of codebase. I can call that function from somewhere else knowing that forceError actually means omitError, but a dev reading only function signature will of course assume the opposite. For example, this function was [called inside the resolveName()](https://github.com/aeternity/aepp-sdk-js/blob/8e872b465b1b782be28babf4c4e9881877bd6d43/es/chain/node.js#L205), and I think the intention here was to force an error throw, but sending true as forceError param was actually silencing the throw. Could you look into this? I noticed this because I was getting weird error "AENS: Invalid name domain. Possible domains [chain]" while doing nothing related to AENS functionality. So while digging through the codebase I found that resolveName() function was not throwing error on assertedType() as it should in my opinion.

davidyuk:
I see your point. I'm fine to merge this if you rename `forceError` to `omitError` and keep the initial logic.

mradkov:
superseded by: #1081

PR #1081, opened 2020-10-26
assertedType: Make the last parameter more obvious

davidyuk:
an alternative to #1079


mradkov (approved):
It's more clearer this way. Thanks.

PR #1083, opened 2020-10-29
Update testnet URL

davidyuk:
https://sdk-testnet.aepps.com looks a bit outdated, should we switch to https://testnet.aeternity.io?


mradkov:
> https://sdk-testnet.aepps.com looks a bit outdated, should we switch to https://testnet.aeternity.io?

Yes, we should. Good catch.

davidyuk:
@mradkov so, let's merge this one

mradkov (approved):

PR #1087, opened 2020-11-12
add vscode .history folder to gitignore

nikita-fuchs:


davidyuk (approved):
Probably this is not a native vscode history, but a third-party history plugin according to https://stackoverflow.com/a/46830780/6176994

If you don't want to mess with it in different projects you can add `.history` to [your global .gitignore file](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c).

Please, setup the support of editorconfig files in your ide, to enable [this setting](https://github.com/aeternity/aepp-sdk-js/blob/0c3cae6f83026b763e57255ec969b4812c17416c/.editorconfig#L8).

Issue #1106, opened 2021-01-11
Problem with node's semantic version digits count


mradkov:
https://forum.aeternity.com/t/the-js-sdk-does-not-support-5-6-3-1/8578

davidyuk:
Semver supports only 3 digits, this problem is on the node side.

PR #1107, opened 2021-01-12
Fix semantic version comparison

markin-io:
A fix for [problem](https://github.com/aeternity/aepp-sdk-js/issues/1106) with node's semantic version digits count


davidyuk (commented):
The code changes look good, but I don't think that this should be fixed because it is obviously a SemVer misuse made on the node side. As I know, it is fixed in 5.6.4 version.

mradkov:
> A fix for [problem](https://github.com/aeternity/aepp-sdk-js/issues/1106) with node's semantic version digits count

As @davidyuk already mentioned this was caused by SemVer misuse from the node side, thus I will close this PR.

Thank you for the contribution, though!

Issue #196, opened 2019-01-04
Simplify package installation and usage


davidyuk:
> **Wait, it just means import and export – not other future JavaScript features?**
Yes. If you're using things like arrow functions and classes (or more exotic features like decorators and object spread), you should transpile them if you want your library to work in environments that don't support those features. Otherwise, whoever uses your library will have to transpile it themselves, including configuring their transpiler to handle your code, which might involve esoteric transforms and conflicting versions of Babel.
It's a frequent source of frustration. Be a responsible library author and ship code that actually runs in the environments you support!

https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features

This article says that tree-shakable code can be delivered in a way that doesn't ask a user to make any additional setup on his side. As I see it is true because all packages that I'm using doesn't require additional setup except JS SDK.

## Expected Behavior
```
import <...> from '@aeternity/aepp-sdk';
// tree-shakable, without changes in Webpack configs
```

## Current Behavior
```
import <...> from '@aeternity/aepp-sdk/es';
```
, required changes in Webpack configs

## Possible Solution
Transpile all es6 features except `import`/`export`

## Possible Implementation
#117

j77ar:
Boot

davidyuk:
Is it closed accidentally?

noandrea:
is this still relevant?

davidyuk:
sure! I hope it would be implemented someday

PR #1113, opened 2021-02-19
Add typescript support for aepp-sdk

the-icarus:
I added basic typescript support for the aepp-sdk so we are able to code in typescript and also users of the sdk get type completion in typescript.

This is an initial pull request which already contains transformation where I already converted `utils/account-formatter.js` and `utils/bignumber.js` to typescript.

There are still open discussions how we want to build the SDK. With the current setup javascript users might only able to use the build assets from `dist` so I guess we still need to adjust the build process but this is something I want to discuss with the current developers/maintainers of the SDK.

Happy to receive your feedback.


the-icarus:
Hi @davidyuk , 
I changed the PR to be a `draft`. 
There is still some things to discuss and to change. I updated the build so during the `npm pack` and `npm publish` phase it will run `tsc` to process everything by the typescript compiler and get type definition files for all source files. Those are added to the `dist/` folder.
I also want to change the package structure. Currently, everything is contained in a release. In my option only the build sources (everything in `dist/`) should be part of a release. This will exclude stuff like tests, raw docs,... from release which will lead to a clean npm package. please share your thoughts on this. As a first step we can compile and publish in a way that we don't break the implementation of our users.

the-icarus:
@davidyuk I made a final change and customized the `npm pack` so the generated code `.js` + `.d.ts` files are at the same location as the original code. This will allow us to add typescript without introducing breaking changes to our users.

Happy to receive your feedback!

the-icarus:
@davidyuk Any update regarding the review?

omar-saadoun:
Hi @davidyuk  please check this

.babelrc
```diff
@@ -1,9 +1,12 @@
...
+    "@babel/preset-typescript"
+  ],
   "plugins": [
     "@babel/plugin-proposal-object-rest-spread",
     "@babel/plugin-transform-runtime",
     "@babel/plugin-proposal-export-default-from",
     "ramda"
   ]
-}
+}
```
davidyuk, line 12: enable support of editorconfig in your ide to make it automatically [insert a final newline](https://github.com/aeternity/aepp-sdk-js/blob/e295ec90fe4bbac30451ddbc40f82e6a3b6e71e4/.editorconfig#L8)

.gitignore
```diff
@@ -21,3 +21,4 @@ coverage.*
 /examples/**/dist
 .history
 .site
+*.tgz
```
davidyuk, line 24: `npm pack` is an edge case, I don't think that ignore files shouldn't be aware of it

.npmignore
```diff
@@ -0,0 +1,2 @@
+es_src
```
davidyuk, line 1: `.npmignore` overrides `.gitignore`, currently it brings extra files to the packed package, for example, I got a local package about 40mb because of `node_modules` of examples. So, if `.npmignore` is used, it should contain all records from `.gitignore`. Alternatively, I'm proposing to use a `files` property of `package.json` to add only necessary folders (`dist` and `es`).

es/utils/amount-formatter.ts
```diff
@@ -56,7 +56,7 @@ export const DENOMINATION_MAGNITUDE = {
  * @param {String} [options.denomination='aettos'] denomination of amount, can be ['ae', 'aettos']
  * @return {String}
  */
-export const toAe = (value, { denomination = AE_AMOUNT_FORMATS.AETTOS } = {}) => formatAmount(value, { denomination, targetDenomination: AE_AMOUNT_FORMATS.AE })
+export const toAe = (value: string | number | BigNumber, { denomination = AE_AMOUNT_FORMATS.AETTOS } = {}) => formatAmount(value, { denomination, targetDenomination: AE_AMOUNT_FORMATS.AE })
```
davidyuk, line 59: this line is too long, I'm planning to limit it by linter in future

es/utils/bignumber.ts
```diff
@@ -3,42 +3,32 @@
  * @module @aeternity/aepp-sdk/es/utils/bignumber
  * @example import { parseBigNumber, asBigNumber, isBigNumber, ceil } from '@aeternity/aepp-sdk/es/utils/bignumber'
  */
-import BigNumber from 'bignumber.js'
+import BigNumber from 'bignumber.js';
```
davidyuk, line 6: `;` is not allowed by our linter rules, need to switch from `standard` to `ts-standard` for linting

package.json
```diff
@@ -12,16 +13,18 @@
...
+    "test:integration:channel,accounts": "nyc --require @babel/register --require ts-node/register mocha './test/integration/@(channel|accounts).js' --exit --timeout 10m",
+    "test:integration:contract,chain": "nyc --require @babel/register --require ts-node/register mocha './test/integration/@(contract|chain).js' --timeout 10m",
+    "test:integration:else": "nyc --require @babel/register --require ts-node/register mocha './test/integration/!(channel|accounts|contract|chain).js' --timeout 10m",
+    "test:unit": "nyc --require @babel/register --require ts-node/register mocha './test/unit/*'",
+    "test:watch": "mocha --recursive --require @babel/register --require ts-node/register --watch --timeout 10m",
     "prepare": "npm run build",
     "genChangelog": "conventional-changelog -p angular -i CHANGELOG.md -s",
-    "prepublishOnly": "npm run docs:docco && npm run docs:api"
+    "prepublishOnly": "npm run docs:docco && npm run docs:api",
+    "prepack": "tsc && node tooling/prepack.js",
```
davidyuk, line 26: proposing to move `tsc` to `build` script

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "@aeternity/aepp-sdk",
   "version": "7.7.0",
   "description": "SDK for the æternity blockchain",
-  "main": "dist/aepp-sdk.js",
+  "main": "dist/index.js",
```
davidyuk, line 5: this file is `es/index.js`, as I understand this change needs to be reverted

package.json
```diff
@@ -64,8 +67,14 @@
     "@babel/plugin-proposal-object-rest-spread": "7.10.1",
     "@babel/plugin-transform-runtime": "7.10.1",
     "@babel/preset-env": "7.10.2",
+    "@babel/preset-typescript": "^7.12.17",
     "@babel/register": "7.10.1",
     "@babel/runtime": "^7.10.2",
+    "@types/bignumber.js": "^5.0.0",
```
davidyuk, line 73: proposing to remove this package because of warning:
> npm WARN deprecated @types/bignumber.js@5.0.0: This is a stub types definition for bignumber.js (https://github.com/MikeMcl/bignumber.js/). bignumber.js provides its own type definitions, so you don't need @types/bignumber.js installed!

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+  "ts-node": {
+    "compilerOptions": {
+      "module": "commonjs"
+    }
+  },
+  "compilerOptions": {
+    "outDir": "./dist/es",
+    "noImplicitAny": true,
+    "module": "es6",
+    "target": "es2017",
```
davidyuk, line 12: switching it to `es5` should solve #196

tooling/prepack.js
```diff
@@ -0,0 +1,5 @@
+const fs = require('fs-extra')
+
+fs.move('es', 'es_src')
```
davidyuk, line 3: Proposing a more radical change: to rename current `es` to `src` (as it should be) and to build typescript straight to `/es`

davidyuk (commented):
All these notices are resolved in #1131. @the-icarus please take a look at that PR.

tooling/prepack.js
nikita-fuchs: +1

davidyuk:
refactored version merged in #1131

PR #1115, opened 2021-02-26
Use swagger-client instead of a custom implementation

davidyuk:
Breaking changes:
- drops `axiosConfig` support because it uses `cross-fetch` that should be lighter than `axios`

Future work:
- write some tests for swagger.js
- investigate it is possible to initialize swagger.js in sync by a prefetched spec file
- generate endpoint handlers on build time to avoid extra request to the node


es/utils/swagger.js
```diff
@@ -16,513 +16,82 @@
...
+ * @return {Object} Swagger client
+ * @example (await genSwaggerClient('https://mainnet.aeternity.io/api')).getAccountByPubkey('ak_jupBUgZNbcC4krDLR3tAkw1iBZoBbkNeShAq4atBtpFWmz36r')
+ */
+export default async (specUrl, internalUrl) => {
+  const spec = await (await fetch(specUrl)).json();
+
+  ['claim', 'preclaim', 'transfer', 'revoke', 'update'].forEach(name => {
+    const s = spec.paths[`/debug/names/${name}`]?.post
+    if (s && !s.consumes) s.consumes = ['application/json']
+  })
```
davidyuk, line 48: this should be removed after resolving https://github.com/aeternity/aeternity/issues/3539

mradkov: Can you log this into an issue, or mark this PR as `wip` until that is fixed?

davidyuk: I have created #1124

mradkov (approved):
LGTM

PR #1117, opened 2021-03-01
[DON'T MERGE] Client side Sophia compiler

gorbak25:
Demo integration with a JS blob containing:
- The entire erlang stdlib
- The entire Fate/AEVM Assembler and Dissasembler
- All data serialization libraries from the node
- The full Sophia compiler(including access to erlang exclusive features as disabling contract hermetization and the ability to generate both the ACI and bytecode at the same time) 

I deliberately didn't do any tree shaking/optimization on the blob. In the future we shall maintain:
- A blob with ALL tools(aestudio etc...)
- A slim calldata encoder/decoder integrated by default in the sdk
- etc...


es/contract/compiler.js
```diff
@@ -29,11 +29,26 @@ import ContractBase from './index'
 import semverSatisfies from '../utils/semver-satisfies'
 import AsyncInit from '../utils/async-init'
 import { VM_TYPE } from '../tx/builder/schema'
+import PS from 'erlps-aesophia/dist/bundle'
+
+const showTerm = PS["Data.Show"].show(PS["Erlang.Type"].showErlangTerm);
+const erlBinToErlStr = (t) => PS["Erlang.Builtins"].erlang__binary_to_list__1([t]);
+const erlStrToJSStr = (t) => PS["Erlang.Type"].fromErl(PS["Erlang.Type"].stringFromErlang)(t).value0
+const jsStrToErlStr = PS["Erlang.Type"].toErl(PS["Erlang.Type"].stringToErlang)
+const mkTuple = PS["Erlang.Type"].ErlangTuple.create;
```
radrow, line 38: `Erlang.Type` exposes nice constructors for these

gorbak25: Nope - you need to pass in the type class dictionary for the inner type.

gorbak25:
Ok the new unoptimized bundle works fine - contracts compile fine using it. The erlps compiler generates the same bytecode as the HTTP compiler.
image
I'm leaving the demo for now. Feel free to try it locally @thepiwo @nikita-fuchs @davidyuk @mradkov :)

es/contract/compiler.js
```diff
@@ -77,13 +93,48 @@ async function validateByteCodeAPI (bytecode, source, options = {}) {
...
+      interface   => StringACI}
+    */
+    const eopts = this.erlpsPrepareCompilerOption(options)
+    const jsonAci = PS['Aeso.Aci'].erlps__contract_interface__3([mkAtom('json'), jsStrToErlStr(code), eopts]).value0[1]
+    const stubAci = PS['Aeso.Aci'].erlps__render_aci_json__1([jsonAci]).value0[1]
+    const ret = { interface: erlStrToJSStr(erlBinToErlStr(stubAci)) }
+    // TODO: implement erlang__atom_to_binary__2
+    // console.log(JSON.parse(erlTermToJSONBIN(json_aci).value0.toString()))
+    console.log(ret)
+    const a = await this.http.post('/aci', { code, options: this.prepareCompilerOption(options) }, options)
```
nikita-fuchs, line 132: heheh, one literally calls an emulated http endpoint ?

nikita-fuchs:
So we `import es/contract/compiler.js`

and call 

compileContractAPI (code, options = {})

and

contractGetACI (code, options = {})

? What are the respective `options` here ?

gorbak25:
> So we `import es/contract/compiler.js`
> 
> and call
> 
> compileContractAPI (code, options = {})
> 
> and
> 
> contractGetACI (code, options = {})
> 
> ? What are the respective `options` here ?

The same as in the original HTTP compiler :P As I said the new compiler should be integrated more appropriately as the options for the HTTP compiler are - Backend, Filesystem, SourceFile - nothing more...
The JS compiler has the following options:
```

-type option() :: pp_sophia_code
                | pp_ast
                | pp_types
                | pp_typed_ast
                | pp_icode
                | pp_assembler
                | pp_bytecode
                | no_code
                | keep_included
                | debug_mode
                | {backend, aevm | fate}
                | {include, {file_system, [string()]} |
                            {explicit_files, #{string() => binary()}}}
                | {src_file, string()}
                | {aci, aeso_aci:aci_type()}.
```

gorbak25:
@nikita-fuchs Treat this PR as an example of using 
import PS from 'erlps-aesophia/dist/bundle'

nikita-fuchs:
sorry I'm not exactly getting it yet 😄 can you provide some example code what to import and what to call ?

gorbak25:
> sorry I'm not exactly getting it yet smile can you provide some example code what to import and what to call ?

Sure! @nikita-fuchs 
In order to access everything you need to ``` npm install erlps-aesophia  ``` - That's the compiler for which a demo integration is shown in this PR.
To access the world of Erlps you then ``` import PS from 'erlps-aesophia/dist/bundle' ```
PS is a map from names to modules which can be called directly from JS :)
Erlscripten maps erlang names as follows:
join(map(capitalize, name.split("_")), ".")
For example the erlang module "aeso_ast_infer_types" is named in erlscripten as "Aeso.Ast.Infer.Types" and you would access it in JS using ```PS['Aeso.Ast.Infer.Types']```
Next the names of erlang functions are prefixed with "erlps__" and sufixed with the function arity. For example if I would like to call "unfold_types_in_type/3" from "aeso_ast_infer_types"(https://github.com/aeternity/aesophia/blob/b8002029cf2510fac61dc7a55b41d03eb22c76ec/src/aeso_ast_infer_types.erl#L17)
You would access this function in JS as ```PS['Aeso.Ast.Infer.Types'].erlps__unfold_types_in_type__3```
Such transpiled erlang functions always take a list of arguments. Examples:
```
PS['Foo'].erlps_demo_fun__0([]) // 0 arity function called with 0 arguments
PS['Foo'].erlps_demo_fun__0([a1]) // 0 arity function called with 1 argument - throws an exception due to arity mismatch
PS['Bar'].erlps_demo_fun__1([a1]) // 1 arity function called with 1 argument 
PS['Baz'].erlps_demo_fun__2([a1, a2]) // 2 arity function called with 2 argument 
```
Transpiled erlang functions always take in erlang terms as arguments - You can create them using the Erlang.Type module, Examples:
```
PS['Erlang.Type'].ErlangTuple.create([]) %% In erlang: {}
PS['Erlang.Type'].ErlangAtom.create("asdf") %% In erlang: asdf
PS['Erlang.Type'].ErlangEmptyList.value %% In erlang: []
PS['Erlang.Type'].ErlangCons.create(PS['Erlang.Type'].ErlangAtom.create("asdf"))(PS['Erlang.Type'].ErlangEmptyList.value) %% In erlang: [asdf]
PS['Erlang.Type'].ErlangTuple.create([PS['Erlang.Type'].ErlangAtom.create("asdf")])%% In erlang: {asdf}
```
There are also some utilities defined to make JS/ERL interoperability easy - but this is WIP still:
```
PS['Erlang.Type'].toErl(PS['Erlang.Type'].stringToErlang)("hello world!") %% In erlang "hello world!"
```
The naming scheme is different for builtin functions - BIF's. If you want to call an erlang BIF then just find it here:
https://github.com/erlscripten/erlps-core/blob/main/src/Erlang/Builtins.purs#L3

Full example:
Let's say I want to compile something simple:
```
// Discover the world of erlang :)
import PS from 'erlps-aesophia/dist/bundle'

// JS string with the contract I want to compile
const demo = "contract Foo =\n    entrypoint init() = ()";

const erlstr = PS['Erlang.Type'].toErl(PS['Erlang.Type'].stringToErlang)(demo);  // The above string as an erlang term - we're now in erlang

// This call returns either {ok, compiled_contract} or a list of errors(or it might throw an exception :P)
const compiled = PS['Aeso.Compiler'].erlps__from_string__2([erlstr, PS['Erlang.Type'].ErlangEmptyList.value]) // 100% equivalent to: aeso_compiler:from_string(erlstr, []).

// Now assuming the compilation worked(returned {ok, _}) we may serialize the code
const compiled1 = PS['Aeser.Contract.Code'].erlps__serialize__1([compiled.value0[1]]) // compiled.value0[1] unpacks the second tuple element from {ok, ...}

// And convert it to cb_****
const bytecode = PS['Aeser.Api.Encoder'].erlps__encode__2([PS['Erlang.Type'].ErlangAtom.create('contract_bytearray'), compiled1])

// Now we have our beloved bytecode ready to be used - we can pass it to other erlang functions or we may just convert the erlang binary to a JS string
// There are better ways to do this conversion but here is one of them :P
const erlBinToErlStr = (t) => PS['Erlang.Builtins'].erlang__binary_to_list__1([t])
const erlStrToJSStr = (t) => PS['Erlang.Type'].fromErl(PS['Erlang.Type'].stringFromErlang)(t).value0
console.log(erlStrToJSStr(erlBinToErlStr(bytecode)))

// And here is how you can pretty print an erlang term:
console.log(PS['Data.Show'].show(PS['Erlang.Type'].showErlangTerm)(bytecode))
```
Here is a screenshot that it works:
image

@nikita-fuchs Does this clear up everything?

es/contract/compiler.js
gorbak25: No it's not emulated - this is only to compare the results of the JS compiler with the HTTP one

nikita-fuchs:
This is absolutely amazing !!

I guess the bytecode is what I pass to the SDKs deployment function then? Let me test this.

es/contract/compiler.js
```diff
@@ -29,18 +29,104 @@ import ContractBase from './index'
...
+        return r;
+    }
+    if(t instanceof erlMap) {
+        t = erlMapToErlList(t);
+        let r = {};
+        while (!(t instanceof erlNil)) {
+            let k = t.value0.value0[0];
+            if (k instanceof erlAtom) k = k.value0;
+            else if (k instanceof erlBin) k = erlBinToJSStr(k);
+            else if (k instanceof erlInt) k = PS["Erlang.Utils"].bigIntToInt(k.value0).value0
```
davidyuk, line 84:
```suggestion
            if (k instanceof erlAtom or k instanceof erlBin or k instanceof erlInt) {
              k = erlJSONToJsJSON(k)
            }
```

es/contract/compiler.js
```diff
@@ -29,18 +29,104 @@ import ContractBase from './index'
...
+        if(t.value0 == "true") return true;
+        if(t.value0 == "false") return false;
+        return t.value0
+    }
+    if(t instanceof erlFloat) return t.value0;
+    if(t instanceof erlInt) {
+      return PS["Erlang.Utils"].bigIntToInt(t.value0).value0
+    };
+    console.log(t)
+    throw new Error("Invalid value")
```
davidyuk, line 106:
```suggestion
    throw new Error(`Invalid value: ${t}`)
```
This exception can be handled by someone without necessity to write to console

davidyuk (commented):
Looks nice, but we need to do something with [bundle size](https://github.com/erlscripten/erlps-aesophia/issues/6) (is it about 7MB now?) also would be nice to have a [high-level interface](https://github.com/erlscripten/erlps-aesophia/issues/7) on the package side

PR #1118, opened 2021-03-03
fix(swagger): workaround to don't break aeternity-encoded keys

davidyuk:
it is made by @CedrikNikita request


es/utils/string.js
```diff
@@ -77,3 +77,5 @@ export function isBase64 (str) {
   const index = str.indexOf('=')
   return !!(index === -1 || str.slice(index).match(/={1,2}/))
 }
+
+export const isAeternityPrefixed = key => key.match(/^[a-z]{2}_/)
```
davidyuk, line 81: Actually, it may break other methods, need to make this check more specific.

PR #1126, opened 2021-03-18
Improve handling of call error

### LINKED ISSUE #1122: artefacts in error decoding
Hey everybody,

```
error : cb_OXJlcXVpcmUgZmFpbGVkarP9mg==
decodedError : 9require failedj���
```

Would it be somehow possible to decode the error in a cleaner fashion, or is it safe to just strip the first and the last couple chars? If not, shall it be fixed rather in the compiler? (cc @hanssv )

davidyuk:
fixes partially #1122

### DIFF
--- a/test/integration/contract.js
+++ b/test/integration/contract.js
@@ -32,6 +32,9 @@ contract Identity =
 const errorContract = `
 contract Identity =
  payable stateful entrypoint main(x : address) = Chain.spend(x, 1000000000)
+
+ payable stateful entrypoint foo() =
+    require(false, "CustomErrorMessage")
 `
 
 const stateContract = `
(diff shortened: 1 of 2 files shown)


test/integration/contract.js
```diff
@@ -371,7 +374,12 @@ describe('Contract', function () {
     try {
       await deployed.call('main', [await contract.address()])
     } catch (e) {
-      e.message.indexOf('Invocation failed').should.not.be.equal(-1)
+      e.message.should.be.equal('Invocation failed')
+    }
+    try {
+      await deployed.call('foo')
+    } catch (e) {
+      e.message.should.be.equal('Invocation failed: "ICustomErrorMessage"')
```
davidyuk, line 382: Still need to figure out how to remove `I` prefix

nikita-fuchs: maybe we can have hans change something about the compiler output here, so you won't have to spend time messing with this prefix issue ?

davidyuk: yep, let's merge it as it is now

mradkov (approved):

PR #1152, opened 2021-04-24
refactor: don't retrieve account from process.env

davidyuk:
SDK is a library and retrieving the account from the process environment seems to be more suitable for the end-user product like sdk cli or so. Also, it may lead to unexpected behavior if sdk would be used in an app that manages accounts by itself.


mradkov (approved):
The code changes seem reasonable. Can you just rebase and resolve the conflicts?

PR #1153, opened 2021-04-24
Update dependencies and don't reexport RLP

davidyuk:


src/utils/crypto.js
```diff
@@ -22,7 +22,7 @@
  */
 
 import bs58check from 'bs58check'
-import * as RLP from 'rlp'
+import { decode as rlpDecode, encode as rlpEncode } from 'rlp'
```
mradkov, line 25: These functions are used as just `decode` and `encode` across other files. Let's keep things consistent and either use them like this everywhere or as `rlpDecode` & `rlpEncode` everywhere?

davidyuk: I've made them consistent.

Originally I was thinking to rename/don't rename depending on the context, like if we have some kind of encode/decode functions except rlp's in this file or not.

Also, I'm rewriting `import * as RLP from 'rlp' ` to specific imports as a good practice to allow three shaking, but actually in this case it doesn't change anything because `rlp` exports just `decode`/`encode` functions.

Motivation to avoid reexporting of the `rlp` module is to
- don't be needed to document it
- makes obviously that this is a third-party implementation, without any modifications on our side

mradkov (approved):

PR #1162, opened 2021-04-27
Base compatibility with Iris node release

### LINKED ISSUE #1160: Ensure IRIS compatibility

davidyuk:
closes #1160 
tested with Iris in https://github.com/aeternity/aepp-sdk-js/pull/1163

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -73,17 +73,33 @@ async function getConsensusProtocolVersion (protocols = [], height) {
  * @example Node({url: 'https://testnet.aeternity.io'})
  */
 const Node = AsyncInit.compose({
-  async init ({ url, internalUrl }) {
+  async init ({ url, internalUrl, ignoreVersion }) {
     if (!url) throw new Error('"url" required')
     this.url = url.replace(/\/$/, '')
     this.internalUrl = internalUrl ? internalUrl.replace(/\/$/, '') : this.url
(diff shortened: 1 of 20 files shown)


src/node.js
```diff
@@ -77,7 +77,11 @@ const Node = AsyncInit.compose({
     if (!url) throw new Error('"url" required')
     this.url = url.replace(/\/$/, '')
     this.internalUrl = internalUrl ? internalUrl.replace(/\/$/, '') : this.url
-    const client = await genSwaggerClient(`${this.url}/api`, { internalUrl: this.internalUrl })
+    let client = await genSwaggerClient(`${this.url}/api`, { internalUrl: this.internalUrl })
+    if (client.spec.info.version.split('-')[0] === '0.0.0') {
```
mradkov, line 81: What shall we do with this `0.0.0`? Iris is planned as `6.0.0` 🤔

davidyuk: I would replace it with something meaningful before the release, currently, it works with that we have 🙃

davidyuk: Actually, I have something to propose, let's increase the upper bound of the node version to < 7.0.0 (implemented in the last commit). Would be nice to merge it because after that we will be able to safely refer to this branch as a part of develop branch.

mradkov (approved):

PR #1174, opened 2021-05-10
chore: avoid ts definitions based on broken JsDoc

davidyuk:


nikita-fuchs:
don't have the rights to approve :/ but will test from PR branch.

PR #1177, opened 2021-05-11
fix: revert conversion of case in calls to compiler

davidyuk:
partially reverts 0d821614e83623f5aeb04133197e8c16de7818b1


thepiwo (approved):
fixes the problems I had calling contracts

mradkov (approved):

PR #1183, opened 2021-05-15
refactor: remove unused asymmetric encode/decode functions

davidyuk:
- it was added in #466 without any reasoning
- it isn't used by sdk itself
- it is unclear who and for what reason can use it
- `nacl.box`/`nacl.box.open` can be used directly instead

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16379,14 +16379,6 @@
       "integrity": "sha512-jtD6YG370ZCIi/9GTaJKQxWTZD045+4R4hTk/x1UyoqadyJ9x9CgSi1RlVDQF8U2sxLLSnFkCaMihqljHIWgMg==",
       "dev": true
     },
-    "ed2curve": {
-      "version": "0.3.0",
-      "resolved": "https://registry.npmjs.org/ed2curve/-/ed2curve-0.3.0.tgz",
-      "integrity": "sha512-8w2fmmq3hv9rCrcI7g9hms2pMunQr1JINfcjwR9tAyZqhtyaMN991lF/ZfHfr5tzZQ8c7y7aBgZbjfbd0fjFwQ==",
-      "requires": {
-        "tweetnacl": "1.x.x"
-      }
-    },
     "electron-to-chromium": {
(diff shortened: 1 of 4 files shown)


mradkov (commented):
I think this is helpful to have in the SDK. Not sure if anyone is and how much they are using it but it gives a useful wrapper for asymmetric encryption.

example case: https://github.com/mradkov/secure-data-exchange

mradkov:
I propose to keep it here. Its obvious that anyone can use it with the underlying `nacl` library directly, but people relying on the sdk to make aeternity related asymmetric crypto operations will expect it to be here.

davidyuk:
Were these functions requested to implement by you? Do you know according to what it is implemented?

 [tweetnacl](https://www.npmjs.com/package/tweetnacl#third-party-libraries-based-on-tweetnacljs) have a reference to https://github.com/whs/tweetnacl-sealed-box that does the same, but a bit more size-efficient, why don't use this instead of our implementation?

I'm not sure what our implementation is audited/well-written, at least the package `ed2curve` (we are using it) says "there's currently no proof that this is safe to do".

I don't like to have this also because other parts of SDK are not compatible with it. It can't be used in aepp, because aepp-wallet connection support only signing.

> https://github.com/mradkov/secure-data-exchange

In your case would be more secure to use exactly `nacl.box` because both Alice and Bob have keypairs (no need to generate ephemeral key pair, sender would be verified despite blockchain).

I'm not sure why are you using `Crypto.decryptData` (based on `nacl.box`) with `Crypto.decryptPrivateKey` (AES). Is it for performance reasons? As I understand `nacl.box` is already using a symmetric cipher underhood (`xsalsa20-poly1305`) that is accessible separately in `nacl.secretbox`.

Simple wrappers won't help to implement crypto-stuff with confidence, better to use crypto libraries directly. The best that we can propose now is a function converting address to byte array.

mradkov:
@davidyuk can you rebase this PR so we can merge it?

mradkov (approved):

PR #1187, opened 2021-05-18
fix(transformation)!: return map instead of array of arrays

davidyuk:
https://github.com/zloirock/core-js/issues/147
polyfills even if they are not necessary wraps native objects, so `instanceof` won't work with maps returned by sdk


davidyuk:
implemented in #1313, will fix polyfilling separatelly

PR #1192, opened 2021-05-22
refactor: drop compatibility with node@5

davidyuk:
I think it is fine to merge after endpoint nodes get updated to 6.0.0 and the hardfork happens on mainnet.


.env
```diff
@@ -1,2 +1,2 @@
-NODE_TAG=v5.10.1
-COMPILER_TAG=v4.3.2
+NODE_TAG=v6.0.0-rc3
```
mradkov, line 1:
```suggestion
NODE_TAG=v6.0.0
```

PR #1195, opened 2021-05-25
Deprecate internal API, switch to protectedDryRunTxs

davidyuk:
uses feature implemented in https://github.com/aeternity/aeternity/pull/3628


docker/aeternity_node_mean16.yaml
```diff
@@ -4,6 +4,9 @@ peers: []
 http:
   external:
     port: 3013
+    gas_limit: 6000000
```
davidyuk, line 7: remove after releasing https://github.com/aeternity/aeternity/pull/3636/files#diff-690f49ae9dfba4976ae5844eca53cdb772079fc2a517d9eab92a1dab0f1d08c7R313

PR #1197, opened 2021-05-26
update compiler version check to lower than v7

nikita-fuchs:


mradkov (commented):
This looks good, but can we maybe think of a `force` option similar to the python SDK so we don't require people to update the SDK every time there is a version bump in the compiler or the node.

Maybe allow users to set the compiler and node versions they want to use in a config or something like this. 🤔 

cc @davidyuk

davidyuk:
@mradkov we already have an `ignoreVersion` [option](https://github.com/aeternity/aepp-sdk-js/blob/58f8908ce08dc1e083e743c463695c24ff6d91ef/src/contract/compiler.js#L61-L65). It was named `forceCompatibility` before, I think it is a super confusing name.

I've added some commits to make tests work fine with compiler@6.

davidyuk (approved):

PR #1206, opened 2021-06-02
Fix lris updating domain name maximum height issue

sunbx:
The validity of domain name renewal after IRIS should be changed to `180000`, currently, it is still `5000`. Restrictions are made in the SDK. I have modified this


mradkov (approved):

davidyuk:
This one break tests in develop branch, also this commit will be ignored by changelog generator 🤷‍♀️

PR #1211, opened 2021-06-06
refactor: validate nodes using Joi

davidyuk:


mradkov (approved):
Much cleaner now, thanks!

PR #1212, opened 2021-06-06
refactor: use joi instead of joi-browser

davidyuk:


src/contract/aci/transformation.js
```diff
@@ -319,14 +318,14 @@ export function transformDecodedData (aci, result, { skipTransformDecoded = fals
...
  * @return {Object} JoiSchema
  */
-export function prepareSchema (type, { bindings } = {}) {
+function prepareSchema (type, bindings) {
   const { t, generic } = readType(type, { bindings })
 
   switch (t) {
     case SOPHIA_TYPES.int:
-      return Joi.number().error(getJoiErrorMsg)
+      return Joi.number().unsafe()
```
mradkov, line 326: `unsafe()` is used because of the BigInt, right?

davidyuk: We have places like https://github.com/aeternity/aepp-sdk-js/blob/a018bf7154e5da1683180bd663ed6f0643efca2a/test/integration/contract.js#L213
As I got, new Joi started to check if number potentially lost precision, this is a good thing but this is a breaking change for us, so I'm disabling it.

davidyuk:
@mradkov it is done, but seems that I can't merge to develop by myself anymore

PR #1213, opened 2021-06-07
refactor: inline unneeded getContractAuthFan helper

davidyuk:


src/contract/ga/helpers.js
```diff
@@ -16,10 +15,4 @@ export const prepareGaParams = (ins) => async (authData, authFnName) => {
   }
 }
 
-export const getContractAuthFan = (ins) => async (source, fnName) => {
-  const { bytecode } = await ins.contractCompile(source)
-
-  return { bytecode, authFun: hash(fnName) }
```
davidyuk, line 22: as I see, it does two unrelated simple things that should be better to inline instead

mradkov: Is it even possible for the `auth` function to be named differently?

davidyuk: yep, it is a part of GA attach transaction: https://github.com/aeternity/aepp-sdk-js/blob/5ad9bafdfd0b544d53664bdcf118a50671dfcb14/src/tx/builder/schema.js#L626

PR #1219, opened 2021-06-16
docs: use custom script instead of broken docco for examples

davidyuk:
`docco` is not working because it ignores `template` options that are not parsed properly because of the new `commander` version provided by the node according to too wide [version range](https://github.com/jashkenas/docco/blob/6dbb7e20096e6778f4c5a5c0896337d2d4cee87f/package.json#L16).

`docco` package seems to be not actively maintained, so I'm proposing to implement own solution.


marc0olo:
will have an immediate look into this. thanks @davidyuk ! :)

marc0olo (commented):
@davidyuk the produced markdown looks like this when I execute `npm run docs:examples`:

```
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
[object Object]
```

are you sure this is working as expected?

marc0olo:
@davidyuk I think we can replace docco with docco-next - at least temporarily. it seems to work fine on my end and produces markdown like I expect it to be including code highlighting.

let's not put further work into this from your side right now. I think your time is better invested in other areas :-)

davidyuk:
> are you sure this is working as expected?

Nope, the last-minute improvement didn't work well 🙃 

> let's not put further work into this from your side right now

I won't, please recheck the current state.

marc0olo:
> I won't, please recheck the current state.

let me first get a workable state on my branch that produces a documentation like we discussed. when this is done I will have a look into this again ;-) ... I definitely welcome dropping the docco stuff which doesn't seem to be maintained very well

marc0olo:
closing this as #1229 is rebased on this branch

PR #1226, opened 2021-06-18
fix: swagger file of aeternity's compiler

### LINKED ISSUE #1225: `contractCompile` using swagger client produces error (`v8.2.0`)
again I am playing around with examples in the docs and recently updated to the release `v8.2.0`

it seems like there was some change that results in `contractCompile` producing sth. which leads to a bad request when calling the http compiler. when using `v8.1.0` the same code works as expected

## Function
```js
await client.contractCompile(CONTRACT_CODE)
```

## Error
```bash
Error: Bad Request
    at http (/home/marc0olo/.nvm/versions/node/v16.3.0/lib/node_modules/@aeternity/aepp-sdk/node_modules/swagger-client/lib/http.js:106:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  status: 400,
  statusCode: 400,
  response: {
    ok: false,
    url: 'https://compiler.aepps.com:443/compile',
    status: 400,
    statusText: 'Bad Request',
    headers: {
      connection: 'close',
      'content-length': '54',
      'content-type': 'application/json',
      date: [ 'Thu', '17 Jun 2021 21:52:00 GMT' ],
      server: 'Cowboy'
    },
    text: '{"info":"","parameter":"","reason":"validation_error"}',
    data: '{"info":"","parameter":"","reason":"validation_error"}',
    body: { info: '', parameter: '', reason: 'validation_error' },
    obj: { info: '', parameter: '', reason: 'validation_error' }
  }
}
```

davidyuk:
fixes #1225

### DIFF
--- a/src/utils/swagger.js
+++ b/src/utils/swagger.js
@@ -46,6 +46,10 @@ export default async (specUrl, { spec, internalUrl, disableBigNumbers, disableCa
   spec = spec || await (await fetch(specUrl)).json()
   const jsonImp = disableBigNumbers ? JSON : JsonBig
 
+  if (new URL(specUrl).origin === 'https://compiler.aepps.com') {
+    spec.schemes = ['https']
+  }
+
   const [external, internal] = await Promise.all([specUrl, internalUrl].map((url) => {
     if (!url) return null
     return SwaggerClient({


src/utils/swagger.js
```diff
@@ -46,6 +46,10 @@ export default async (specUrl, { spec, internalUrl, disableBigNumbers, disableCa
   spec = spec || await (await fetch(specUrl)).json()
   const jsonImp = disableBigNumbers ? JSON : JsonBig
 
+  if (specUrl.startsWith('https://compiler.aepps.com')) {
+    spec.schemes = ['https']
+  }
+
```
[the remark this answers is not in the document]
davidyuk: Using git blame you would found this PR and the corresponding issue where I have explained it.

src/utils/swagger.js
```diff
@@ -46,6 +46,10 @@ export default async (specUrl, { spec, internalUrl, disableBigNumbers, disableCa
   spec = spec || await (await fetch(specUrl)).json()
   const jsonImp = disableBigNumbers ? JSON : JsonBig
 
+  if (new URL(specUrl).origin === 'https://compiler.aepps.com') {
```
mradkov, line 49: What if one is not using the hosted compiler?

Is the utils/swagger only supposed to work with the hosted one?

marc0olo: good point. can we just check if the url starts with `https://`?

nikita-fuchs:
are we continuing ? ;)

src/utils/swagger.js
davidyuk: I believe this is an issue of compiler deployment because if it deployed over HTTPS and have `"schemes":["http"]` in a swagger file, then the swagger file is incorrect 🤷‍♀️ 

Our compiler is also deployed improperly, but I believe it would be fixed soon and we will be able to remove this.

mradkov (approved):

PR #1229, opened 2021-06-22
Docs improvements

### LINKED ISSUE #1220: docs: cleanup and improvements
I am working on this right now as discussed. this is required prior to including the docs as submodule in the documentation hub, see https://github.com/aeternity/documentation-hub/issues/220

we have some javascript examples that are documented and converted via docco into markdown. as this is generally a great thing but doesn't fit our needs I have temporarily switched to docco-next. when I am done with the cleanup I will revisit https://github.com/aeternity/aepp-sdk-js/pull/1219 and we should be able to get rid of the docco stuff.

marc0olo:
closes #1220


marc0olo:
the current state uses Material docs and we should keep in mind that we want to include the docs in the documentation-hub, see https://github.com/aeternity/documentation-hub/issues/220. so if it is included in the documentation-hub as submodule it doesn't matter what theme is defined here because the documentation hub will define it.

@davidyuk I am not sure what's wrong with linting right now. for some reason the pipeline fails. maybe it's due to changes in the examples folder which are used to generate the markdown files. if so we should ignore that. can you check that?

this PR also includes changes from #1219

@nikita-fuchs how is https://aeternity-js-sdk.readthedocs.io/ being served right now? could we publish the state of this there? I am thinking about how to deal with "standalone" docs if we want to include everything in the documentation hub.

marc0olo:
currently there are github actions in place that seem to build the docs and publish the docs to gh-pages. is this working? @davidyuk 

what das following action do?
```yaml
      - uses: peaceiris/actions-gh-pages@v3
        if: github.ref == 'refs/heads/master'
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: docs
```

when I try to access gh-pages I always get redirected to aeternity.com. do we have some policy in place that performs this redirect on company level?

marc0olo:
I need to provide some small changes to the github action regarding the docs. playing around on my personal account right now.

the current state of the docs can be previewed here:
- https://marc0olo.github.io/aepp-sdk-js/

marc0olo:
I added a dedicated github action for generating the docs that is currently running for pushes to `develop` and `docs-improvements`.

each update to the `docs-improvements` branch will now result in updates to https://aeternity.com/aepp-sdk-js/

I am wondering how we should proceed here in the future. maybe we should trigger the docs action manually for specific branches if necessary. this way we would be flexible to display a certain state of the documentation in case of fixes.

the-icarus:
Would it make sense to have two instances one would show the current develop and one would show the docs for the releases version?

Actually, I don't see a reason to show docs for a non-releases version of the SDK.

marc0olo:
> Would it make sense to have two instances one would show the current develop and one would show the docs for the released version?
> 
> Actually, I don't see a reason to show docs for a non-releases version of the SDK.

generally agree here. the thing is that at least this state doesn't publish the API reference anymore (temporarily) and that we might add some examples and guides (#1243, #1244) in the short term which could be independent from any release.

personally I would like to get the docs generally cleaned up first before introducing versions based on releases.

it shouldn't be that hard to introduce versioning for mkdocs-material (https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/) when we have a clean setup. here some examples how that could look like:
- https://squidfunk.github.io/mkdocs-material-example-versioning

marc0olo:
making this a draft again. looking into versioning right now. I am using a plugin called `mike` which is included in MkDocs material by default.

I was playing around on my own fork and somehow it seems like the plugin always pushed to the `origin` instead of my personal upstream. opened an issue for that => https://github.com/jimporter/mike/issues/62

so we have some strange commits to the gh-pages branch here, sorry for that! 😅

marc0olo:
versioning is now active.

@mradkov unfortunately I deleted the gh-pages branch and this probably caused the repo to not serve the docs anymore. can you take a look at the settings?

after this PR is merged following things should be in place for the docs:
- `develop` docs are updated on each merge to the `develop` branch (currently the only version available)
- docs for each release tag are published when a release is made
   - the most recent release tag gets the `latest` alias which is set and served by default

mradkov (approved):

PR #1250, opened 2021-07-08
fix: swagger https issue

mradkov:
Added `https` check for the origin URL of the compiler, not only for the hosted compilers.

workaround for https://github.com/aeternity/aesophia_http/issues/70

ref: https://github.com/aeternity/aepp-sdk-js/issues/1225#issuecomment-863985169


marc0olo (approved):
we should make sure that the next release of the http compiler fixes https://github.com/aeternity/aesophia_http/issues/70 so that we can drop that workaround

Issue #1252, opened 2021-07-08
`TxBuilderHelper.getMinimumNameFee` returns wrong value for names with length >= 31


marc0olo:
**Describe the bug**
See issue title. I am playing around with `AENS` delegations and already raised a question regards to BigNumber, see #1251. It seems like the SDK currently returns an invalid value for the smallest possible name fee.

Currently the method described above returns a `BigNumber` with the value `2865700000000000000`.

**Expected behavior**
The method should return the correct value `3000000000000000000`, see https://github.com/aeternity/protocol/blob/master/AENS.md#protocol-fees-and-protection-times

**Please tell us about your environment:**

  - Node Version: v6.1.0
  - Protocol Version: 5
  - SDK Version: v8.2.1

PR #1254, opened 2021-07-13
Fix name length minimum bid fee

mradkov:
fixes: #1252

### DIFF
--- a/src/tx/builder/schema.js
+++ b/src/tx/builder/schema.js
@@ -36,10 +36,11 @@ export const NAME_MAX_CLIENT_TTL = 84600
 export const CLIENT_TTL = NAME_MAX_CLIENT_TTL
 export const NAME_FEE = 0
 // # see https://github.com/aeternity/aeternity/blob/72e440b8731422e335f879a31ecbbee7ac23a1cf/apps/aecore/src/aec_governance.erl#L67
-export const NAME_FEE_MULTIPLIER = 100000000000000
+export const NAME_FEE_MULTIPLIER = 1e14 // 100000000000000
 export const NAME_FEE_BID_INCREMENT = 0.05 // # the increment is in percentage
 // # see https://github.com/aeternity/aeternity/blob/72e440b8731422e335f879a31ecbbee7ac23a1cf/apps/aecore/src/aec_governance.erl#L272
 export const NAME_BID_TIMEOUT_BLOCKS = 480 // # ~1 day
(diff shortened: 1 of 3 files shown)


src/tx/builder/schema.js
```diff
@@ -36,11 +36,11 @@ export const NAME_MAX_CLIENT_TTL = 84600
 export const CLIENT_TTL = NAME_MAX_CLIENT_TTL
 export const NAME_FEE = 0
 // # see https://github.com/aeternity/aeternity/blob/72e440b8731422e335f879a31ecbbee7ac23a1cf/apps/aecore/src/aec_governance.erl#L67
-export const NAME_FEE_MULTIPLIER = 100000000000000
+export const NAME_FEE_MULTIPLIER = 1e14 // 100000000000000
 export const NAME_FEE_BID_INCREMENT = 0.05 // # the increment is in percentage
 // # see https://github.com/aeternity/aeternity/blob/72e440b8731422e335f879a31ecbbee7ac23a1cf/apps/aecore/src/aec_governance.erl#L272
 export const NAME_BID_TIMEOUT_BLOCKS = 480 // # ~1 day
-export const NAME_BID_MAX_LENGTH = 12 // # this is the max length for a domain to be part of a bid
+export const NAME_BID_MAX_LENGTH = 31 // # this is the max length for a domain to be part of a bid
```
marc0olo, line 43: is this used somewhere in the tests? I think this constant relevant to know if a claim results in an auction or not, see
image

marc0olo (changes requested):
I guess `NAME_BID_MAX_LENGTH` should be reverted and is correct. not sure if it is used internally.

src/tx/builder/schema.js
marc0olo: ok I just see that it was used in the method `getMinimumFee`. I am maximally confused right now. need to double check 😅

mradkov: It is used internally, reverting it back to 12 breaks the fee calculation again as it is used in here: https://github.com/aeternity/aepp-sdk-js/blob/51e3ac50594b97fb1fa6511057821e268d461f83/src/tx/builder/helpers.js#L291

And yes, I think you are correct this will break the bidding. I'm fixing it now by introducing another const which will be `NAME_MAX_LENGTH_FEE = 31` which will be indicating the max length of a name that requires a base fee to be paid.

marc0olo: yeah the logic you adapted is correct whereas the variable names are confusing

PR #1258, opened 2021-07-15
docs: add 9.0.0 migration guide

davidyuk:
The next time needs to try to put it into `BREAKING CHANGE` footer. Probably then it will automatically appear in release notes.

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer


marc0olo:
@mradkov looks good to me know. I made a few small adjustments

mradkov (approved):

PR #1262, opened 2021-07-28
Enable commitmentHash preclaim in tests

mradkov:


test/integration/contract.js
```diff
@@ -215,12 +215,11 @@ describe('Contract', function () {
 
       // preclaim
       const { salt: _salt } = await contract.aensPreclaim(name)
```
marc0olo, line 217: the `aensPreclaim` should be replace with the contract call and delegation signature.

test/integration/contract.js
```diff
@@ -215,12 +215,11 @@ describe('Contract', function () {
...
-      // @TODO enable after next HF
-      // const commitmentId = commitmentHash(name, _salt)
+      const commitmentId = commitmentHash(name, _salt)
```
marc0olo, line 218: the salt used here needs to be generated with the utils independent from the `aensPreclaim` method

marc0olo (changes requested):

davidyuk (approved):

marc0olo:
@mradkov I made some mistake. fixed it and squashed to commits into one. it's unfortunate to not see tests failing due to that linting and codecov errors

mradkov (commented):
LGTM now

PR #1268, opened 2021-09-01
Add events package to package manager file to fix webpack building

subhod-i:
Since latest webpack versions dropped the default "events" package, during a fresh setup of the SDK manual installation of the "events" package is necessary. Saved it in the `package.json` file


mradkov (changes requested):
Please rename the commit message to follow the convetional commits standard.

See lint error here:
https://github.com/aeternity/aepp-sdk-js/runs/3483515684#step:8:8

mradkov (approved):

Issue #1224, opened 2021-06-17
account is required to use the `Universal` Stamp of the SDK


marc0olo:
while working on #1220 I noticed that it seems like we currently need to provide an account to the `Universal` Stamp. if no account is provided an error occurs.

following example currently doesn't work and produced error message `Data doesn't match expected type ak`
```js
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
</head>
<body>
  <!-- include latest SDK version -->
  <script src="https://unpkg.com/@aeternity/aepp-sdk/dist/aepp-sdk.browser-script.js"></script>
  <script type="text/javascript">
    Ae.Node({ url: 'https://testnet.aeternity.io' }).then(node => {
        Ae.Universal({
              nodes: [{ name: 'testnet', instance: node }]
            }).then(aeInstance => {
              aeInstance.height().then(height => {
                console.log("Current Block Height:" + height)
              })
            })
    })
  </script>
</body>
</html>
```

marc0olo:
@mradkov this should definitely be included in the next release, too

subhod-i:
The above example works with the changes in the attached PR.

Output if you run the above code
```
No account/accounts supplied
Current Block Height:485196
```

PR #1270, opened 2021-09-02
Fix default account required to setup universal stamp of the SDK issue

subhod-i:
Issue descriptions:  #1224 

Process accounts in `accounts stamp` only when it's not empty.
Modify BaseAe SDK setup in integration tests to accomodate accountless instances.
Use Accountless stamp(openClient) in Chain tests.

### DIFF
--- a/src/account/multiple.js
+++ b/src/account/multiple.js
@@ -51,13 +51,13 @@ import AccountBase, { isAccountBase } from './base'
 export default AccountBase.compose(AsyncInit, {
   async init ({ accounts = [], address }) {
     /* An Account maynot be required for the Node/Chain methods */
-    if(Array.isArray(accounts) && accounts.length) {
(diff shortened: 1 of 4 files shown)


subhod-i:
Typo in the commit message. Fixing it.

subhod-i:
Fixed it.

mradkov:
Is this still a draft?

subhod-i:
Yes. There are few more places in tests which can use accountless stamps. Right now accounless-stamps are used in just `chain tests`

mradkov (commented):
This only fixes the integration tests, what has been done to the BaseAe stamp in order to fix the actual issue?

Ahh I see, so there is an additional check to verify the accounts is not empty and only then proceed. Ok, seems like it works.

mradkov (approved):

PR #1273, opened 2021-09-08
Modify default gas to a reasonable number and update docs with Gas estimation

### LINKED ISSUE #1267: reconsider default gas limit / improve docs for contract related tx's
## Problem
For whatever reason there has been a default gas limit of `1_579_000` defined in the past, see https://github.com/aeternity/aepp-sdk-js/blob/5f1cb1105eff0393b4cd45c96a88202698013ec1/src/tx/builder/schema.js#L25

Though each transaction only uses the gas it really consumes and users/devs can always adjust that limit, the current default value is completely unnecessary and plain wrong.

Assuming that most developers don't adjust that default gas limit and just use the default value this leads to the fact that the protocol at the moment (can be improved but requires hardfork, see https://github.com/aeternity/aeternity/issues/3706) only includes 3 of such transactions into a microblock. Only `6_000_000` gas fits into one microblock.

## Solution
We should at least decrease the default gas limit to a bare minimum. Personally I am even thinking about removing that default value completely. Users (at least devs) should know how to deal with that topic and if we try to "hide complexity" in that regards most of the (unexperienced) folks doesn't care about that at all and just relies on the default values all the time.

## Update documentation
We should highlight this topic more in the documentation and give advise how to deal with gas estimation and gas prices.

Here a comment from @hanssv in that regards:
> Until the network gets more crowded gas price isn’t very interesting - but it is good to be prepared.

> For gas limit I’m no crypto-app developer but I’d envisioned you would go about it something like this!? I’m hacking away at my idea, eventually I end up with a contract and some code around that. The app will typically call a couple of endpoints in the contract - I would then make a bunch of test cases that exercise these with (somewhat) realistic data. I would run the tests and see how much gas is needed. If the contract is well designed the gas needed should be reasonably stable if not it doesn’t scale and you have other issues further along… This should give me a ballpark figures for my contract calls and I can go about and multiply them by 1.5 and deploy my app. Then you’d probably want to measure and monitor your app. dry-run is also an option but most projects shouldn’t need to do that. At the end of the day you only pay for the gas you consume so a small over-estimate is the way to go.

And additional comments of @dincho:
> I would use x2 :)

> In regards to gas price the node can provide some basic and poor estimate just like geth, if developers need more accurate estimates they use some other service

> But still a poor estimate provided by the node would improve the current situation drastically IMO
That estimate can be done client side as well at the cost of performance

subhod-i:
Issue disucssion: Fix #1267 

Solution: Change default gas from ` 1600000 - 21000` to `25000`.  Discussion on the new value https://github.com/aeternity/aepp-sdk-js/issues/1267#issuecomment-909291108 

Updated the docs to reflect the changes.

### DIFF
--- a/docs/transaction-options.md
+++ b/docs/transaction-options.md
@@ -85,7 +85,7 @@ The following options are sepcific for each tx-type.
     - You can specify the denomination of the `amount` that will be provided to the contract related transaction.
 
 ## How to estimate gas?
-- As æpp developer, It is reasonable to estimate the gas consumption for a contract call using the dry-run feature of the node **once** and provide a specific offset (multiplied by 1.5 or 2) as default in the æpp to ensure that contract calls are mined. Depending on the logic of the contract the gas consumption of a specific contract call can vary and therefore you should monitor the gas consumption and increase the default for the respective contract call accordingly.
+- As æpp developer, it is reasonable to estimate the gas consumption for a contract call using the dry-run feature of the node **once** and provide a specific offset (e.g. multiplied by 1.5 or 2) as default in the æpp to ensure that contract calls are mined. Depending on the logic of the contract the gas consumption of a specific contract call can vary and therefore you should monitor the gas consumption and increase the default for the respective contract call accordingly over time.
 - The default `gas` value of `25000` covers trivial contract calls but, The recommended way is estimating the gas using the dry-run feature instead of relying on the default value.
 
 
--- a/docs/transaction-options.md
+++ b/docs/transaction-options.md
@@ -46,8 +46,8 @@ The following options are sepcific for each tx-type.
     - To be used for providing `aettos` (or `AE` with respective denomination) to a contract related transaction.
 - `denomination` (default: `aettos`)
     - You can specify the denomination of the `amount` that will be provided to the contract related transaction.
-- `gas` (default: `1600000 - 21000`)
-    - Max. amount of gas to be consumed by the transaction.
+- `gas` (default: `25000`)
+    - Max. amount of gas to be consumed by the transaction. Learn more on [How to estimate gas?](#how-to-estimate-gas)
 - `gasPrice` (default: `1e9`)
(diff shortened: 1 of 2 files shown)


docs/transaction-options.md
```diff
@@ -82,4 +82,10 @@ The following options are sepcific for each tx-type.
 
 ### SpendTx
 - `denomination` (default: `aettos`)
-    - You can specify the denomination of the `amount` that will be provided to the contract related transaction.
\ No newline at end of file
+    - You can specify the denomination of the `amount` that will be provided to the contract related transaction.
+
+## How to estimate gas?
+- As æpp developer, It is reasonable to estimate the gas consumption for a contract call using the dry-run feature of the node **once** and provide a specific offset (multiplied by 1.5 or 2) as default in the æpp to ensure that contract calls are mined. Depending on the logic of the contract the gas consumption of a specific contract call can vary and therefore you should monitor the gas consumption and increase the default for the respective contract call accordingly.
```
marc0olo, line 88:
```suggestion
- As æpp developer, it is reasonable to estimate the gas consumption for a contract call using the dry-run feature of the node **once** and provide a specific offset (e.g. multiplied by 1.5 or 2) as default in the æpp to ensure that contract calls are mined. Depending on the logic of the contract the gas consumption of a specific contract call can vary and therefore you should monitor the gas consumption and increase the default for the respective contract call accordingly over time.
```

docs/transaction-options.md
```diff
@@ -82,4 +82,10 @@ The following options are sepcific for each tx-type.
...
+- The default `gas` value of `25000` covers trivial contract calls but, The recommended way is estimating the gas using the dry-run feature instead of relying on the default value.
```
marc0olo, line 89:
```suggestion
- The default `gas` value of `25000` should cover all trivial contract calls. In case transactions start running out of gas you should proceed the way described above and estimate the required gas using the dry-run feature.
```

mradkov (approved):

Issue #1242, opened 2021-06-28
refactor & fix: oracle delegation signature


marc0olo:
this is related to #1237 

-------------------------

there are two methods that generate the same signature and can be used to `register` and `extend` an oracle on behalf of an account:

```js
await client.delegateOracleRegisterSignature(contractAddress, {onAccount})

await client.delegateOracleExtendSignature(contractAddress, {onAccount})
```

additionally there is another method to create signatures that allow the contract to respond to a certain query which currently **does NOT work**!

there is an open TODO in the integration test section that should cover this for 17 months now 😞 - see:
https://github.com/aeternity/aepp-sdk-js/blob/9c4a49d6ffa2ecbc7c390aabab8c87eca62e0924/test/integration/contract.js#L270

```js
await client.delegateOracleRespondSignature(oracleId, queryId, contractAddress, { onAccount })
```

the reason that it doesn't work is that under the hood each of these methods call following method:
```js
async function delegateSignatureCommon (ids = [], opt = {}) {
  return this.sign(
    Buffer.concat(
      [
        Buffer.from(this.getNetworkId(opt)),
        decode(await this.address(opt)),
        ...ids.map(e => decode(e))
      ]
    ),
    opt
  ).then(s => Buffer.from(s).toString('hex'))
}
```

the method `delegateSignatureCommon` **always** includes the `network_id` and `address` of the account but according to the Sophia documentation (which is correct according to @hanssv) there is no need to include the `address` in the delegation signature that is used for oracles to respond to a certain query. the delegation signature used for oracles to respond instead must contain following attributes:
- `network id` + `oracle query id` + `contract address`

--------------------------------------
identified during the work on #1220

subhod-i:
@marc0olo  Not sure if I have caught the definition correctly.

The `delegateSignatureCommon` method shouldn't include `address` for signing but instead, the format should be
```javascript
async function delegateSignatureCommon (ids = [], opt = {}) {
  return this.sign(
    Buffer.concat(
      [
        Buffer.from(this.getNetworkId(opt)), // Network Id
        decode(this.getContractAddress(opt)), // Contract address
        Buffer.from(this.getOracleQueryId(opt)),  // Oracle query Id
        ...ids.map(e => decode(e))
      ]
    ),
    opt
  ).then(s => Buffer.from(s).toString('hex'))
}
```

marc0olo:
there are two types of things addressed here:
[...]

Issue #1237, opened 2021-06-23
refactor confusing methods for AENS delegation signature creation


marc0olo:
I don't understand why we have 3 different methods that all generate the same delegation signature.

## current methods
```js
let sig;
const contractId = 'ct_7UfopTwsRuLGFEcsScbYgQ6YnySXuyMxQWhw6fjycnzS5Nyzq'
const name = 'abc.chain'

// name preclaim signature delegation
sig = await client.delegateNamePreclaimSignature(contractId)
console.log(sig)

// name claim signature delegation
sig = await client.delegateNameClaimSignature(contractId, name)
console.log(sig)

// name transfer signature delegation
sig = await client.delegateNameTransferSignature(contractId, name)
console.log(sig)

// name revoke signature delegation
sig = await client.delegateNameRevokeSignature(contractId, name)
console.log(sig)

/*
b95489afa3f7e5e9fc2ccc7927817ea12e08e28076433c9f76c52a759930029203045417e2cf25f36d9f5a693a677b3ef503082d17b6a7623fe45f6c18ea9e0a
06e15fd4ba83891f830f96049cf6166d5ddc82fdb82db8171fe0b1614b778fac347eea8af33066fa843f225a1383e0dde5ddcb1b2cdeeb91cdc0f9a7dbc33309
06e15fd4ba83891f830f96049cf6166d5ddc82fdb82db8171fe0b1614b778fac347eea8af33066fa843f225a1383e0dde5ddcb1b2cdeeb91cdc0f9a7dbc33309
06e15fd4ba83891f830f96049cf6166d5ddc82fdb82db8171fe0b1614b778fac347eea8af33066fa843f225a1383e0dde5ddcb1b2cdeeb91cdc0f9a7dbc33309
*/
```

## proposed methods
```js
let sig;
const contractId = 'ct_7UfopTwsRuLGFEcsScbYgQ6YnySXuyMxQWhw6fjycnzS5Nyzq'
const name = 'abc.chain'

sig = await client.createAensDelegationSignature(contractId) // defaults to pre-claim signature if no name is provided
console.log(sig)

sig = await client.createAensDelegationSignature(contractId, name)
console.log(sig)

/*
b95489afa3f7e5e9fc2ccc7927817ea12e08e28076433c9f76c52a759930029203045417e2cf25f36d9f5a693a677b3ef503082d17b6a7623fe45f6c18ea9e0a
[...]

PR #1274, opened 2021-09-15
Fix AENS and Oracle delegation signature

subhod-i:
Oracle delegation signature issue: #1242 
Aens delegation signature issue: #1237 

Design & solution: All the AENS and Oracle delegation methods use the common method `delegateSignatureCommon` for signing. The common method used to include the address for signing even if the `onAccount` is not set in the options.  Hence the solution is to modify the common method to use the address for signing only if the `onAccount` key exists in the options.

This will resolve @marc0olo comment https://github.com/aeternity/aepp-sdk-js/issues/1242#issuecomment-918335331 on `delegateOracleRespondSignature`

Further updates on the changes:
* Javascript doesn't support function overloading. The new approach is `Accept params as Object` to merge the common signature delegation methods
* Update AENS & Oracle documentation
* Update AENS and Oracle delegation tests to incorporate new implementation
* Use a different account in Oracle tests to overcome the register oracle conflict over an account

### DIFF
--- a/docs/guides/aens.md
+++ b/docs/guides/aens.md
@@ -114,7 +114,7 @@ Note:
     - It's still possible to pass it as additional param, see [transaction options](../transaction-options.md#nameclaimtx).
 - In case the `claim` triggers an auction the required `nameFee` is locked by the protocol.
     - If you win the auction the `nameFee` is permanently deducted from your accounts balance and effectively *burned*.
-        - It will be credited to `ak_11111111111111111111111111111111273Yts` which nobody can access. This reduces the total supply of AE over time. 
+        - It will be credited to `ak_11111111111111111111111111111111273Yts` which nobody can access. This reduces the total supply of AE over time.
     - If somebody else outbids you the provided `nameFee` is immediately released and returned to your account.
 
 ### Bid during an auction
@@ -347,7 +347,7 @@ console.log(nameRevokeTx)
 Note:
 
 - On revocation the name enters in a `revoked` state.
-- After a timeout of `2016` keyblocks the name will be available for claiming again. 
+- After a timeout of `2016` keyblocks the name will be available for claiming again.
 
 ## Delegate signature to contract (AENS interface)
 It is possible to authorize a Sophia contract to manage an AENS name on behalf of your account. In order to achieve that you need to provide a delegation signature to the contract. The contract will then be able to use the [AENS interface](https://github.com/aeternity/aesophia/blob/v6.0.1/docs/sophia.md#aens-interface) and perform AENS related actions on behalf of your account.
@@ -363,20 +363,13 @@ const client = await Universal({ ... }) // init instance of Universal Stamp
 const contractId = 'ct_asd2ks...'
 // AENS name
 const name = 'example.chain'
+// Sign with a specific account
+const onAccount = await client.address()
 
 // this signature will allow the contract to perform a pre-claim on your behalf
-const sig = await client.delegateNamePreclaimSignature(contractId)
+const sig = await client.createAensDelegationSignature(contractId, { onAccount })
 
 // this signature will allow the contract to perform
 // any name related transaction for a specific name that you own
-const sig = await client.delegateNameClaimSignature(contractId, name)
+const sig = await client.createAensDelegationSignature(contractId, name, { onAccount })
 ```
-
-Note:
-
-- If required you can pass an additional options object like `{ onAccount: <account> }` to these functions in order to sign with a specific account.
-- The SDK currently also provides two other functions to generate a delegation signature:
-    - `delegateNameTransferSignature`
-    - `delegateNameRevokeSignature`
-- Both of these functions produce the same signature as `delegateNameClaimSignature` which can be confusing. There is an open issue that proposes to change that:
-    - https://github.com/aeternity/aepp-sdk-js/issues/1237
--- a/docs/guides/aens.md
+++ b/docs/guides/aens.md
@@ -367,9 +367,9 @@ const name = 'example.chain'
 const onAccount = await client.address()
 
 // this signature will allow the contract to perform a pre-claim on your behalf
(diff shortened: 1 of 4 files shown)


marc0olo:
@subhod-i can you please make sure that this test is now working with your changes and comment it in?
https://github.com/aeternity/aepp-sdk-js/blob/9c4a49d6ffa2ecbc7c390aabab8c87eca62e0924/test/integration/contract.js#L270

and I am still thinking about how we proper solve that scenario that we have multiple different methods that all generate the same signature. we should definitely harmonize that. it confuses devs because they think they need to create different signatures for different actions.

I described this again here:
- https://github.com/aeternity/aepp-sdk-js/issues/1237#issuecomment-919443025

marc0olo:
and of course make sure to update the docs properly ;-)

https://aeternity.com/aepp-sdk-js/develop/guides/aens.html#delegate-signature-to-contract-aens-interface

subhod-i:
Got it.  Action items from the discussions.

- Remove redundant delegation signature methods(AENS + Oracle)
- Enable test https://github.com/aeternity/aepp-sdk-js/blob/9c4a49d6ffa2ecbc7c390aabab8c87eca62e0924/test/integration/contract.js#L270 
- Update Docs to reflect the new implementation strategy https://aeternity.com/aepp-sdk-js/develop/guides/aens.html#delegate-signature-to-contract-aens-interface

marc0olo:
> Got it. Action items from the discussions.
> 
> * Remove redundant delegation signature methods(AENS + Oracle)
> * Enable test https://github.com/aeternity/aepp-sdk-js/blob/9c4a49d6ffa2ecbc7c390aabab8c87eca62e0924/test/integration/contract.js#L270
> * Update Docs to reflect the new implementation strategy https://aeternity.com/aepp-sdk-js/develop/guides/aens.html#delegate-signature-to-contract-aens-interface

@subhod-i what about 1 and 3?

subhod-i:
> > Got it. Action items from the discussions.
> > 
> > * Remove redundant delegation signature methods(AENS + Oracle)
> > * Enable test https://github.com/aeternity/aepp-sdk-js/blob/9c4a49d6ffa2ecbc7c390aabab8c87eca62e0924/test/integration/contract.js#L270
> > * Update Docs to reflect the new implementation strategy https://aeternity.com/aepp-sdk-js/develop/guides/aens.html#delegate-signature-to-contract-aens-interface
> 
> @subhod-i what about 1 and 3?

@marc0olo  This PR is getting big. Hence unlinked the AENS issue #1237 and thus will be creating a new PR for AENS refactoring.

marc0olo:
> @marc0olo This PR is getting big. Hence unlinked the AENS issue #1237 and thus will be creating a new PR for AENS refactoring.

mhh - for signature creation I also have mentioned methods that create the same signature in #1242. is this PR really so big if you fix both at once? doesn't seem to be a problem for me. but if you prefer we can tackle the signature creation refactoring in another PR. from my side it's not needed to separate that

test/integration/contract.js
```diff
@@ -217,26 +217,26 @@ describe('Contract', function () {
       const { salt: _salt } = await contract.aensPreclaim(name)
       // @TODO enable after next HF
       // const commitmentId = commitmentHash(name, _salt)
-      const preclaimSig = await contract.delegateNamePreclaimSignature(contractAddress, { onAccount: current })
+      const preclaimSig = await contract.createAensDelegationSignature(contractAddress, { onAccount: current })
```
marc0olo: no need to distinguish between different signatures here in this test. currently we have:
- preclaimSig
- claimSig
- transferSig
- revokeSig

after your changes we only need following signatures in the test:
- preclaimSig
- aensDelegationSig

marc0olo (changes requested):

test/integration/contract.js
subhod-i: Done. `revokeSig` is a different signature as it needs to be signed by the new owner once the transfer has occurred.
So we do have 3 signatures
* preclaimSig
* aensDelegationSig
* revokeSig

docs/guides/aens.md
```diff
@@ -363,20 +363,13 @@ const client = await Universal({ ... }) // init instance of Universal Stamp
...
 
 // this signature will allow the contract to perform
 // any name related transaction for a specific name that you own
-const sig = await client.delegateNameClaimSignature(contractId, name)
+const aensDelegationSig = await client.createAensDelegationSignature({ contractId, name }, { onAccount })
 ```
-
-Note:
-
-- If required you can pass an additional options object like `{ onAccount: <account> }` to these functions in order to sign with a specific account.
```
marc0olo, line 377: as you removed this note => is not possible to specify another account for signature creation anymore? otherwise we should keep this specific note :)

and I think it would be good to have it working this way, as almost every function allows that

subhod-i: Yes, Absolutely possible to specify another account.

former `delegateSignatureCommon`:  Select one existing account from the stamp to be included in the signature.
new  `delegateSignatureCommon`:  Don't choose any accounts by default unless  `{ onAccount }` in the options.

If you observe the new guide https://github.com/aeternity/aepp-sdk-js/blob/252f5d32d913176a6d0d6c929b3665e4dfcd612f/docs/guides/aens.md#delegate-signature-to-contract-aens-interface It's a must to have `onAccount` option in the methods. It will eliminate accidentally signing the signature from unintended accounts because you need to specifically pass an account field in the options. This may be a limitation given almost every function allows without selecting a specific account. Well, In this case, I have to modify the `delegateSignatureCommon` to choose the default account.

marc0olo: ok, got it

marc0olo (approved):
looks good to me now, thanks @subhod-i !

marc0olo:
@mradkov can you also have a look into this? and we were/are also wondering about that codecov check again 😅

mradkov (approved):
LGTM

Issue #1143, opened 2021-04-05
highlevel postQuery doesn't consider oracle queryFee


thepiwo:
highlevel postQuery should default to it's `oracle.queryFee` as default

thepiwo:
the function postQuery of the higher level sdk oracle functions

PR #1294, opened 2021-10-12
Add Oracle queryfee to the postQuery method 

subhod-i:
Fix Issue #1143 

Solution: Get the Oracle instance and merge them with options. 
Options priority: `Options` from the method parameter > Oralce instance own properties > SDK defaults

### DIFF
--- a/test/integration/oracle.js
+++ b/test/integration/oracle.js
@@ -89,9 +89,9 @@ describe('Oracle', function () {
     const account = generateKeyPair()
 
     before(async function () {
-      await client.spend(1e15, account.publicKey)
-      client.addAccount(MemoryAccount({ keypair: account }), { select: true })
-      oracleWithFee = await client.registerOracle("{'city': str}", "{'tmp': num}", { queryFee, onAccount: account })
+      await sdk.spend(1e15, account.publicKey)
+      sdk.addAccount(MemoryAccount({ keypair: account }), { select: true })
+      oracleWithFee = await sdk.registerOracle("{'city': str}", "{'tmp': num}", { queryFee, onAccount: account })
     })
 
     it('Post Oracle Query with default query fee', async () => {


src/ae/oracle.js
```diff
@@ -177,7 +177,8 @@ async function registerOracle (queryFormat, responseFormat, options = {}) {
  * @return {Promise<Object>} Query object
  */
 async function postQueryToOracle (oracleId, query, options = {}) {
-  const opt = R.merge(this.Ae.defaults, options)
+  const oracle = await this.api.getOracleByPubkey(oracleId)
+  const opt = R.mergeAll([this.Ae.defaults, oracle, options])
```
davidyuk, line 181:
```suggestion
  const opt = { ...this.Ae.defaults, ...oracle, ...options }
```

test/integration/oracle.js
```diff
@@ -81,4 +82,31 @@ describe('Oracle', function () {
...
     response.decode().toString().should.be.equal(queryResponse)
   })
+
+  describe('Oracle query fee settings', async () => {
+    let oracleWithFee
+    const queryFee = 24000
+    const account = generateKeyPair()
+
+    before(async function () {
+      await client.spend('1' + '0'.repeat(15), account.publicKey)
```
davidyuk, line 92:
```suggestion
      await client.spend(1e15, account.publicKey)
```
won't it work the same?

src/ae/oracle.js
```diff
@@ -177,7 +177,8 @@ async function registerOracle (queryFormat, responseFormat, options = {}) {
  * @return {Promise<Object>} Query object
  */
 async function postQueryToOracle (oracleId, query, options = {}) {
-  const opt = R.merge(this.Ae.defaults, options)
+  const oracle = await this.api.getOracleByPubkey(oracleId)
```
davidyuk, line 180:
```suggestion
  const queryFee = options.queryFee  || (await this.api.getOracleByPubkey(oracleId)).queryFee
```
then it won't fetch oracle if `queryFee` is provided

PR #1295, opened 2021-10-12
test: name sdk instance consistent

davidyuk:
In different files we have tests of the same SDK's configuration, naming it differently in different files is confusing to me.


davidyuk:
this is a separate topic, do we have any specific questions to this PR?

mradkov (approved):

PR #1301, opened 2021-10-14
Refactor contracts

davidyuk:
I don't know the intention behind `contract.methods.init`, but I think would be better to use `contract.deploy` directly


marc0olo:
I think we should collect feedback from the community (discord/forum) before merging this. not sure if this is heavily being used by somebody.

all in all I see now problem in allowing to call `init` explicitely for deployments.

thepiwo:
Keeping `init` is more logical to me. As a contract dev I write my entrypoint init() as any other method, why not call it as any other method?

marc0olo:
I personally would also prefer having direct access to the methods of the contract interface like:

```js
const myContract = await aeSdk.getContractInstance(...);
await myContract.init([]);
await myContract.doSth();
```

but we leave like is now and don't change as we aren't aligned here. there is some refactoring required and this could have been done simultaneously

PR #1305, opened 2021-10-20
Refactor contracts

davidyuk:
continuation of #1301


mradkov (approved):
I like that we gradually remove R usage. Are those wrappers not used anywhere else?

Otherwise looks good to me.

PR #1311, opened 2021-10-27
Fix Channel tests

subhod-i:
The following tests used to fail due to short polling.

- `can solo close a channel` 
-  `can dispute via slash tx`

Fix: both the tests use `sendTransaction` method with default poll blocks of height `+10`. Change it to 20 so that transaction confirmation will wait till 20 more blocks.

### DIFF
--- a/test/integration/channel.js
+++ b/test/integration/channel.js
@@ -734,7 +734,7 @@ describe('Channel', function () {
       payload: signedTx
     })
     const closeSoloTxFee = unpackTx(closeSoloTx).tx.fee
-    await initiator.sendTransaction(await initiator.signTransaction(closeSoloTx), { waitMined: true })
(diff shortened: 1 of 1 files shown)


subhod-i:
> > No, we haven't. Dev mode has issues with respect to await height. I guess this might be due to instant mining
> 
> what kind of issue did we experience there? we should analyze this independent of this issue. we need to be able to explain in detail what we think is not working there

I will log them here. Have we decided on using dev mode for the SDK tests?

davidyuk (approved):

mradkov (approved):

marc0olo:
> Have we decided on using dev mode for the SDK tests?

not yet. but IMO devmode should behave the same in terms of SDK interaction. but I am not sure if devmode is mature enough yet to integrate into the testsuite

Issue #1290, opened 2021-10-11
rethink & consolidate usage of BigInt / BigNumber throughout the SDK


marc0olo:
known related issues / PRs:
- https://github.com/aeternity/aepp-sdk-js/issues/1251
- https://github.com/aeternity/aepp-sdk-js/pull/1259

we should maybe also consult with @dincho here.

davidyuk:
Some time before I was investigating the ability to implement a custom wrapper around bigint to represent ae coints, but this feature on TS side is missed https://github.com/microsoft/TypeScript/issues/2361

davidyuk:
Approximate roadmap (each step may be done separately):
- [ ] don't use strings as numbers
- [ ] use BigInt instead of BigNumber (drop BigNumber completely)
- [ ] implement Ae wrapper around BigInt

I've touched on this topic in #1762 but it is too heavy for sdk@13

marc0olo:
almost forgot about this one 😅 good to see it being tackled 👍

davidyuk:
Optionally we can have a fancy way to define ae amounts
```
const aeAmount = ae`42.4`;
const fee = aettos`19320000000000`;
```
Also, may be useful to define something between `ae` and `aettos`, a multiple of `MIN_GAS_PRICE` (1e9). Because in the protocol amounts less than `MIN_GAS_PRICE` don't make sense.

dincho:
> Some time before I was investigating the ability to implement a custom wrapper around bigint to represent ae coints, but this feature on TS side is missed [microsoft/TypeScript#2361](https://github.com/microsoft/TypeScript/issues/2361)

That's the right way to go IMO

PR #1313, opened 2021-10-27
Integrate calldata package, drop joi package

### LINKED ISSUE #435: Find an alternative of "joi-browser" package
We were updating SDK of Base app and have noticed a new "joi-browser" package in our dependencies. It consumes ~160kb of bundle size that is probably because it uses by in a way that is can't be tree-shaked or it is not tree-shakable at all. Until SDK is not highly dependent on it, I propose to investigate the ability to tree shake it or find an alternative package.

### LINKED ISSUE #1253: support `AENS` related Sophia types
it seems like the following types are currently not supported:
- https://github.com/aeternity/aesophia/blob/master/docs/sophia_stdlib.md#name
- https://github.com/aeternity/aesophia/blob/master/docs/sophia_stdlib.md#pointee

I tested following method to get an `AENS.name`:
```sophia
    entrypoint get_name(name : string) : option(AENS.name) =
        AENS.lookup(name)
```

I also tried to add a pointer using this method but it seems like `AENS.pointee` isn't supported, too:
```sophia
    stateful entrypoint add_key(addr: address, name: string, key: string,
                                  pt : AENS.pointee, sig : signature) =
        switch(AENS.lookup(name))
            None => ()
            Some(AENS.Name(_, _, ptrs)) =>
                AENS.update(addr, name, None, None, Some(ptrs{[key] = pt}), signature = sig)
```

@davidyuk do we have integration tests where providing these types as param or receiving them as result is covered? I don't think so. at least I am not able to call these methods using the latest version of the SDK ( `v8.2.1`)

I was also testing this with the java implementation that @mitch-lbw and I work on and raised a similar issue there, see https://github.com/kryptokrauts/contraect-maven-plugin/issues/65

### LINKED ISSUE #1312: Replace http compiler with calldata-lib
Related:
https://github.com/aeternity/aesophia/issues/351
https://github.com/aeternity/aepp-calldata-js/pull/72
https://github.com/aeternity/aepp-calldata-js/pull/71
https://github.com/aeternity/aepp-calldata-js/issues/73
https://github.com/aeternity/aepp-calldata-js/issues/74
https://github.com/aeternity/aepp-calldata-js/pull/76

https://github.com/aeternity/aepp-calldata-js/pull/78
https://github.com/aeternity/aepp-calldata-js/pull/79
https://github.com/aeternity/aepp-calldata-js/pull/80
https://github.com/aeternity/aepp-calldata-js/pull/82

https://github.com/aeternity/aepp-calldata-js/pull/84
https://github.com/aeternity/aepp-calldata-js/pull/85
https://github.com/aeternity/aepp-calldata-js/pull/86

davidyuk:
closes #1312, closes #435, closes #1253


thepiwo:
trying to encode `undefined` for an `option(int)` seems to fail with

`Error: Variant value should be an object with "variant" and "values" fields, got undefined instead`

marc0olo:
@thepiwo We have two open issues in the calldata lib regarding variants:
- https://github.com/aeternity/aepp-calldata-js/issues/73
- https://github.com/aeternity/aepp-calldata-js/issues/74

Not sure how this is related. But just to let you know.

test/integration/contract.js
```diff
@@ -319,33 +320,34 @@ describe('Contract', function () {
...
     expect(result.result).to.be.equal(undefined)
     result.txData.should.not.be.equal(undefined)
     await sdk.poll(result.hash, { interval: 50, attempts: 1200 })
   })
 
   it('calls deployed contracts static', async () => {
-    const result = await deployed.callStatic('getArg', ['42'])
-    return result.decode().should.eventually.become(42)
+    const result = await deployed.callStatic('getArg', [42])
+    expect(result.decodedResult).to.be.equal(42n)
```
thepiwo, line 342: why is this with 'n' here?

marc0olo: it's `BigInt` which is used in the calldata lib right now. this is also sth. that should be addressed with the integration. I have some open issues in that regards, see:
- https://github.com/aeternity/aepp-sdk-js/issues/1290

thepiwo: we should figure this out or provide a different solution with chai if this is the issue

davidyuk:
@mradkov I think this one is ready to merge

test/integration/contract.js
marc0olo: @davidyuk can you verify the problem and comment this again?

davidyuk: I've created a new project with the latest mocha (9.1.3) and chai (4.3.4):
```js
import { expect } from 'chai'

it('works', () => {
    expect({ a: 1n }).to.be.eql({ a: 2n })
})
```
`mocha test.js` outputs:
```
  1) works:

      AssertionError: expected { a: 1n } to deeply equal { a: 2n }
      + expected - actual

       {
      -  "a": "1"
      +  "a": "2"
       }
```
in the diff section at the end BigInts presented as strings for some reason. If this is the issue, I don't think that we should do anything on our side to avoid that.

davidyuk:
```js
expect({ a: '1' }).to.be.eql({ a: 2n })
```
```
       {
      -  "a": "1"
      +  "a": "2"
       }
```
🤷‍♀️

thepiwo: I still don't get why we have to have it like this. If it stays like this there needs to be a big explainer everywhere for ALL aepp devs to adjust their aepps

thepiwo:
`map()` return types seem to always be empty when decoded

thepiwo (approved):
pretty speedy

test/integration/contract.js
marc0olo: you mean the `BigInt` in general? well I think in the past we just weren't able to use it and in the calldata lib it makes definitely sense to use it as it is possible now.

I can understand your point of view. but isn't `BigInt` the right way to do it?

@thepiwo wouldn't you have to touch all repos again anyway with the upcoming AEproject release or how will that transition look like again?  (can't exactly remember)

I mean it's definitely a bit unfortunate but it will be a major release - so we could (and maybe should) do it this way 🤷‍♀️

thepiwo: No, repos that don't use aeproject don't have to be touched, e.g. governance-server, governance-aepp, superhero wallet, graffiti-aepp, graffiti-server.

We should just make the use of BigInt very clear in docs and Changelogs and so on, I'm not saying its bad to have it.

PR #1316, opened 2021-11-03
feat(poll-interval): reduce poll interval to be a more sensible default

thepiwo:


marc0olo:
I saw some discussion about that in the chat. how should we proceed here? @davidyuk @mradkov @subhod-i

thepiwo:
I don't think there is a better proposal yet, we can keep the bad default we have now and delay EVERY TX action by 2-5sec depending on use-case while waiting for someone to come around with a better solution, I know how to fix it for stuff that affects me

mradkov:
> I don't think there is a better proposal yet, we can keep the bad default we have now and delay EVERY TX action by 2-5sec depending on use-case while waiting for someone to come around with a better solution, I know how to fix it for stuff that affects me

Agree here. Lets merge this and refine later in case it causes issues.

mradkov (approved):

PR #1320, opened 2021-11-08
perf: take current height from status response

davidyuk:


mradkov (approved):

davidyuk:
@mradkov Is code coverage a required check? Maybe I can't merge because of it?

Issue #1256, opened 2021-07-14
`callStatic` as function of the contract instance


marc0olo:
my current assumption is that the contract instance has the method `callStatic` when an actual deployment is being performed using the instance while `callStatic` is missing when getting the contract instance for an already deployed contract.

while `call` can be used for dry-run also when providing options `{callStatic: true}`.

I am 100% sure I was able to call `callStatic` on the instance when I updated the docs.

now I am confused in that regards. if we always want devs to use `call` while providing `callStatic: true` in the options we should remove `callStatic` method completely from the instance.

I see no reason why the SDK adds this method when deploying the contract if we don't add it when initializing the instance for a specific contract address.

I see two options:
1. never add `callStatic` to the contract instance (force the devs to use `call` with the respective option)
     - this requires an update of the documentation as there it's documented that `callStatic` can be used directly
1. also add `callStatic` to the contract when initializing it for a specific contract address
     - would allow us to keep the documentation as is

mradkov:
Yeah, I think this should be refactored if it does not imply any other usage elsewhere.

marc0olo:
what option should we choose? removing it or make sure it's always present in the instance?

marc0olo:
depending on the decision the docs need to be adjusted => https://aeternity.com/aepp-sdk-js/develop/guides/contracts.html#b-regular-entrypoints

mradkov:
IMO removing the `callStatic` fn will be better and only allowing access via the `{ callStatic: true }` option. But lets get @davidyuk's opinion as well.

davidyuk:
I can't find `callStatic` method of contract instance. `call`/`callStatic` can't do more than `methods.*` so I'm for dropping `call` and not implementing `callStatic`.

marc0olo:
@davidyuk are you sure it doesn't exist? when I tested in the past it was added to the contract instance after I deployed a contract.

when I only initialized the contract instance at a specific address without deploying it wasn't existent. that confused my somehow.

can you doublecheck this?

davidyuk:
Ok, I got it. Confusion is between the result of `getContractInstance` and `deployInfo` that is returned by `contractInstance.deploy` (also `sdk.contractDeploy`):
https://github.com/aeternity/aepp-sdk-js/blob/72907f351579444ebd269668f254d4956657e38b/src/contract/aci/index.js#L189-L192
[...]

PR #1329, opened 2021-11-18
Update contract docs

davidyuk:
related to #1256

### DIFF
--- a/docs/guides/contracts.md
+++ b/docs/guides/contracts.md
@@ -189,9 +189,9 @@ Sometimes you might wonder how to pass params to the JavaScript method that call
 | bytes     | `get_bytes(test: bytes(3))`       | `get_bytes(new Uint8Array([0x1, 0x1f, 0x10]))`   |
 | string    | `hello_world(say_hello: string)`  | `hello_world('Hello!')`                          |
 | list      | `have_a_few(candy: list(string))` | `have_a_few(['Skittles', 'M&Ms', 'JellyBelly'])` |
-| tuple     | `a_few_things(things: (string * int * map(address, bool)))` | `a_few_things(['hola', 3, {'ak_1337...': true}])` |
+| tuple     | `a_few_things(things: (string * int * map(address, bool)))` | `a_few_things(['hola', 3, new Map([['ak_1337...', true]])])` |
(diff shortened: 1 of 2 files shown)


docs/guides/contracts.md
```diff
@@ -78,11 +78,10 @@ Note:
         - Key-value map with name of the include as key and source code of the include as value.
     - `validateByteCode` (default: false)
         - Compare source code with on-chain version.
-    - `opt` (default: {})
```
marc0olo, line 81: so `opt` not an explicit property anymore?

docs/guides/contracts.md
```diff
@@ -78,11 +78,10 @@ Note:
...
-        - Object with other [transaction options](../transaction-options.md) which will be provided to **every transaction** that is initiated using the contract instance. You should be aware that:
-            - For most of these additional options it doesn't make sense to define them at contract instance level.
-            - You wouldn't want to provide an `amount` to each transaction or use the same `nonce` which would result in invalid transactions.
-            - For options like `ttl` or `gasPrice` it does absolutely make sense to set this on contract instance level.
+    - other [transaction options](../transaction-options.md) which will be provided to **every transaction** that is initiated using the contract instance. You should be aware that:
```
marc0olo, line 81: options just embedded as object, right? should maybe mentioned a bit more explicit here

marc0olo, line 81:
```suggestion
    - an object including other [transaction options](../transaction-options.md) which will be provided to **every transaction** that is initiated using the contract instance. You should be aware that:
```

davidyuk: there was an embedded object at `opt` field (`opt` inside `options`), but now it is inlined into `options` object.

davidyuk: there is no an object at the end of options, transactions options should be fields in `options` object

marc0olo: I might lack of javascript knowledge here. so having an `options` object as last param of a function call is usual and typically known to devs? I exactly meant what you wrote as answer. I just wanted to be explicit with the `options` object.

I edited the suggestion. is that better? or can we really leave like is?

mradkov (approved):

docs/guides/contracts.md
mradkov: Looks good to me like this.

marc0olo: ahh you mentioned this in the call, right? I am fine with keeping it like is. no big deal :-)

davidyuk: > `options` object

is explained above, don't think necessary to repeat it 🙂

Issue #1332, opened 2021-11-24
inconvenient usage of FATE variant types related to AENS


marc0olo:
I am not sure where to place this issue and how to address this. this issue is closely related to https://github.com/aeternity/aepp-sdk-js/issues/1253 which is now solved and yes, we are able to handle those types now in the SDK.

as @thepiwo recently updated AEproject I took the chance to revisit my old draft PR where I wanted to test all the delegation signature stuff and opened a bunch of issues, see https://github.com/aeternity/aepp-sophia-examples/pull/67

actually I finally succeeded in encoding/decoding AENS relevant stuff (yeah 😎) but I am not really happy in the way we have to deal with the pointers right now:

**js test**
```js
  it('should add and get pointers correctly', async () => {
      const accountPointerKey = SCHEMA.NAME_ID_KEY['ak'];
      const oraclePointerKey = SCHEMA.NAME_ID_KEY['ok'];
      const contractPointerKey = SCHEMA.NAME_ID_KEY['ct'];
      const channelPointerKey = SCHEMA.NAME_ID_KEY['ch'];
      let getPointersResult = await contract.methods.get_pointers(testName);
      console.log(getPointersResult.decodedResult);
      await contract.methods.add_key(nameOwnerKeypair.publicKey, testName, accountPointerKey, {'AENS.AccountPt':[nameOwnerKeypair.publicKey]}, aensDelegationSignature, { onAccount: delegateKeypair });
      await contract.methods.add_key(nameOwnerKeypair.publicKey, testName, oraclePointerKey, {'AENS.OraclePt':[nameOwnerKeypair.publicKey]}, aensDelegationSignature, { onAccount: delegateKeypair });
      await contract.methods.add_key(nameOwnerKeypair.publicKey, testName, contractPointerKey, {'AENS.ContractPt':[nameOwnerKeypair.publicKey]}, aensDelegationSignature, { onAccount: delegateKeypair });
      await contract.methods.add_key(nameOwnerKeypair.publicKey, testName, channelPointerKey, {'AENS.ChannelPt':[nameOwnerKeypair.publicKey]}, aensDelegationSignature, { onAccount: delegateKeypair });
      let aensName = await contract.methods.get_name(testName);
      console.log(aensName.decodedResult);
      pointers = await contract.methods.get_pointers(testName);
      console.log(pointers.decodedResult);
  });
```

**test output**
```sh
Map(0) {}
{
  'AENS.Name': [
[...]

PR #1334, opened 2021-11-24
feature(aens)!: don't limit pointer keys

### LINKED ISSUE #1240: AENS pointers: allow custom keys
Although the protocol allows to set different pointers for each valid type (e.g. it is possible to set keys for multiple accounts) the SDK currently doesn't provide a possibility to do so.

Currently it is only possible to provide an array of values (e.g. `['ak_...', 'ct_...', 'ok_...', 'ch_...']`) where the respective `default key` is determined automatically.

I propose to change that so that developers can (no-breaking change) **OR** have to (breaking change) provide a key-value map with valid types as values.

I think @hanssv said that there is a limit of maximum pointers that can be set. Not sure about the maximum amount right now.

davidyuk:
closes #1240

### DIFF
--- a/src/tx/builder/schema.js
+++ b/src/tx/builder/schema.js
@@ -43,7 +43,7 @@ export const NAME_BID_TIMEOUT_BLOCKS = 480 // # ~1 day
 // # this is the max length for a domain that requires a base fee to be paid
 export const NAME_MAX_LENGTH_FEE = 31
 export const NAME_BID_MAX_LENGTH = 12 // # this is the max length for a domain to be part of a bid
-export const NAME_ID_KEY = {
+export const POINTER_KEY_BY_PREFIX = {
   ak: 'account_pubkey',
   ok: 'oracle_pubkey',
   ct: 'contract_pubkey',
--- a/docs/guides/aens.md
+++ b/docs/guides/aens.md
@@ -176,15 +176,24 @@ Now that you own your AENS name you might want to update it in order to:
 
 ### Set pointers & update TTL
 ```js
+import { getDefaultPointerKey } from '@aeternity/aepp-sdk'
+
 const name = 'testNameForTheGuide.chain'
-const pointersArray = ['ak_2519mBsgjJEVEFoRgno1ryDsn3BEaCZGRbXPEjThWYLX9MTpmk'] // 'ct_...', 'ok_...', 'ch_...'
+const oracle = 'ok_2519mBsgjJEVEFoRgno1ryDsn3BEaCZGRbXPEjThWYLX9MTpmk'
+const pointers = {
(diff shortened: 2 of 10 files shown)


src/tx/builder/schema.js
```diff
@@ -42,12 +42,6 @@ export const NAME_FEE_BID_INCREMENT = 0.05 // # the increment is in percentage
 export const NAME_BID_TIMEOUT_BLOCKS = 480 // # ~1 day
 export const NAME_MAX_LENGTH_FEE = 31 // # this is the max length for a domain that requires a base fee to be paid
 export const NAME_BID_MAX_LENGTH = 12 // # this is the max length for a domain to be part of a bid
-export const NAME_ID_KEY = {
```
marc0olo, line 45: not sure if we need to remove those keys as these are the default keys. maybe we could have a helper function instead to get the default key for the respective type.

marc0olo (commented):
got a question/suggestions regarding the default pointer keys.

could you please also add a test where a custom pointer key independent of the default keys is used? and e.g. a set of all default pointers including the custom pointer is being set and validated?

src/tx/builder/schema.js
davidyuk: I've reverted classify function that is doing that: https://github.com/aeternity/aepp-sdk-js/blob/db6ca4c3d89eedf7fa326bd6d0484ea0d94de19e/src/tx/builder/helpers.js#L251-L269

marc0olo: mhh, I don't think we are talking about the same here. not sure. let me ask a question like this:

- how does the sdk user know about the default pointer-keys that have a special meaning? now he would need to know them, right? shouldn't there be an easy way to look those keys with special meaning up?

it's only about those 4 keys

marc0olo: take a look at this example to know what I mean:
- https://github.com/aeternity/aepp-sdk-js/issues/1332#issue-1061841739

davidyuk: > how does the sdk user know about the default pointer-keys that have a special meaning?

I think the best here is to have it documented. As a developer, I see that each pointer has a name, and looks reasonable to go to documentation to check the default pointer names.

`classify` can be used this way:
```js
const ct = 'ct_.....';
sdk.aensUpdate(name, { [classify(ct)]: ct, myPointer: ct });
```

marc0olo: @davidyuk I think we should rename `classify` if it is only used for this purpose. I would prefer sth. like `getDefaultPointerKey` but I am also open for other proposals. I think a general helper function should be more expressive.

I would also expect this to be in the usage guide for the default pointers and additionally 1 custom pointer to make the usage more clear with some small description

marc0olo (approved):
looks good to me know, thanks @davidyuk! :)

mradkov (approved):

PR #1339, opened 2021-11-26
Introduce error types

### LINKED ISSUE #1194: Introduce different error types
For example, to be able to do
```js
error instanceof NotEnoughBalanceError
```
instead of https://github.com/aeternity/superhero-wallet/commit/d7ee86729ec803fea26531d972033855c72c282c#diff-ead5c468661f39fe883b9b2c6f3c4f34d88d09b7ba32b47adbb056bd9ef1d75bR23-R24

subhod-i:
fix #1194 

An initial template design with custom error instances.


src/utils/error.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+ *  OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ *  PERFORMANCE OF THIS SOFTWARE.
+ */
+
+export class AeError extends Error {}
+
+export class InvalidKeypairError extends AeError { }
+
+export class UnAvailableAccountError extends AeError {
```
mradkov, line 22:
```suggestion
export class UnavailableAccountError extends AeError {

```

src/account/multiple.js
```diff
@@ -134,17 +135,17 @@ export default AccountBase.compose(AsyncInit, {
...
     _resolveAccount (account) {
       if (account === null) {
-        throw new Error('No account or wallet configured')
+        throw new UnknownAccountError()
       } else {
         switch (typeof account) {
           case 'string':
             decode(account, 'ak')
-            if (!this.accounts[account]) throw new Error(`Account for ${account} not available`)
+            if (!this.accounts[account]) throw new UnAvailableAccountError(account)
```
mradkov, line 143:
```suggestion
            if (!this.accounts[account]) throw new UnavailableAccountError(account)
```

src/account/multiple.js
```diff
@@ -123,7 +124,7 @@ export default AccountBase.compose(AsyncInit, {
      */
     selectAccount (address) {
       decode(address, 'ak')
-      if (!this.accounts[address]) throw new Error(`Account for ${address} not available`)
+      if (!this.accounts[address]) throw new UnAvailableAccountError(address)
```
mradkov, line 127:
```suggestion
      if (!this.accounts[address]) throw new UnavailableAccountError(address)
```

src/account/multiple.js
```diff
@@ -25,6 +25,7 @@ import AsyncInit from '../utils/async-init'
 import MemoryAccount from './memory'
 import { decode } from '../tx/builder/helpers'
 import AccountBase, { isAccountBase } from './base'
+import { UnAvailableAccountError, UnknownAccountTypeError, UnknownAccountError } from '../utils/error'
```
mradkov, line 28:
```suggestion
import { UnavailableAccountError, UnknownAccountTypeError, UnknownAccountError } from '../utils/error'
```

test/integration/accounts.js
```diff
@@ -22,6 +22,7 @@ import { generateKeyPair } from '../../src/utils/crypto'
 import BigNumber from 'bignumber.js'
 import MemoryAccount from '../../src/account/memory'
 import { AE_AMOUNT_FORMATS } from '../../src/utils/amount-formatter'
+import { UnAvailableAccountError, UnknownAccountTypeError, UnknownAccountError } from '../../src/utils/error'
```
mradkov, line 25:
```suggestion
import { UnavailableAccountError, UnknownAccountTypeError, UnknownAccountError } from '../../src/utils/error'
```

test/integration/accounts.js
```diff
@@ -159,22 +160,22 @@ describe('Accounts', function () {
...
     it('Fail on invalid account', async () => {
       return expect(sdk.spend(1, await sdk.address(), { onAccount: 1 }))
-        .to.be.rejectedWith('Unknown account type: number (account: 1)')
+        .to.be.rejectedWith(UnknownAccountTypeError, 'Unknown account type: number (account: 1)')
     })
 
     it('Fail on non exist account', async () => {
       return expect(sdk.spend(1, await sdk.address(), { onAccount: 'ak_q2HatMwDnwCBpdNtN9oXf5gpD9pGSgFxaa8i2Evcam6gjiggk' }))
-        .to.be.rejectedWith('Account for ak_q2HatMwDnwCBpdNtN9oXf5gpD9pGSgFxaa8i2Evcam6gjiggk not available')
+        .to.be.rejectedWith(UnAvailableAccountError, 'Account for ak_q2HatMwDnwCBpdNtN9oXf5gpD9pGSgFxaa8i2Evcam6gjiggk not available')
```
mradkov, line 168:
```suggestion
        .to.be.rejectedWith(UnavailableAccountError, 'Account for ak_q2HatMwDnwCBpdNtN9oXf5gpD9pGSgFxaa8i2Evcam6gjiggk not available')
```

mradkov (changes requested):

src/utils/error.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+export class UnknownAccountError extends AeError {
+  constructor () {
+    super('No account or wallet configured')
+  }
+}
+export class UnknownAccountTypeError extends AeError {
+  constructor (account: any) {
+    super(`Unknown account type: ${typeof account} (account: ${String(account)})`)
```
davidyuk, line 34:
```suggestion
    super(`Unknown account type: ${typeof account} (account: ${account})`)
```
won't it work the same?

src/account/multiple.js
```diff
@@ -134,17 +135,17 @@ export default AccountBase.compose(AsyncInit, {
...
           case 'string':
             decode(account, 'ak')
-            if (!this.accounts[account]) throw new Error(`Account for ${account} not available`)
+            if (!this.accounts[account]) throw new UnAvailableAccountError(account)
             return this.accounts[account]
           case 'object':
             return isAccountBase(account) ? account : MemoryAccount({ keypair: account })
           default:
-            throw new Error(`Unknown account type: ${typeof account} (account: ${account})`)
+            throw new UnknownAccountTypeError(account)
```
davidyuk, line 148: it looks like a development-only error, maybe we can use a more general error class:
```suggestion
            throw new TypeError(`Account should be an address (ak-prefixed string), keypair, or instance of account base, got ${account} instead`)
```

src/utils/error.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * ISC License (ISC)
+ * Copyright 2018 aeternity developers
```
davidyuk, line 3:
```suggestion
 * Copyright 2021 aeternity developers
```
It would be more correct. Actually, I didn't see similar each-file licenses in other aeternity projects, maybe we can drop it?

src/account/memory.js
```diff
@@ -49,16 +50,16 @@ export default AccountBase.compose({
       return
     }
 
-    if (!keypair || typeof keypair !== 'object') throw new Error('KeyPair must be an object')
-    if (!keypair.secretKey || !keypair.publicKey) throw new Error('KeyPair must must have "secretKey", "publicKey" properties')
-    if (typeof keypair.publicKey !== 'string' || keypair.publicKey.indexOf('ak_') === -1) throw new Error('Public Key must be a base58c string with "ak_" prefix')
+    if (!keypair || typeof keypair !== 'object') throw new InvalidKeypairError('KeyPair must be an object')
+    if (!keypair.secretKey || !keypair.publicKey) throw new InvalidKeypairError('KeyPair must must have "secretKey", "publicKey" properties')
+    if (typeof keypair.publicKey !== 'string' || keypair.publicKey.indexOf('ak_') === -1) throw new InvalidKeypairError('Public Key must be a base58c string with "ak_" prefix')
```
davidyuk, line 55: Later I would propose to accept only private key here, then we don't need most of this validation, the interface would be simpler

src/account/multiple.js
```diff
@@ -134,17 +135,17 @@ export default AccountBase.compose(AsyncInit, {
      */
     _resolveAccount (account) {
       if (account === null) {
-        throw new Error('No account or wallet configured')
+        throw new UnknownAccountError()
```
davidyuk, line 138: The logic here is broken by previous developers, if I accidentally do
```
sdk.someMethod({ onAccount: null })
```
then SDK would say that I don't have accounts at all?

src/utils/error.ts
```diff
@@ -0,0 +1,36 @@
...
+export class AeError extends Error {}
+
+export class InvalidKeypairError extends AeError { }
+
+export class UnAvailableAccountError extends AeError {
+  constructor (address: string) {
+    super(`Account for ${address} not available`)
+  }
+}
+export class UnknownAccountError extends AeError {
```
davidyuk, line 27:
```suggestion

export class UnknownAccountError extends AeError {
```
let's add an empty line between them

src/utils/error.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  PERFORMANCE OF THIS SOFTWARE.
+ */
+
+export class AeError extends Error {}
+
+export class InvalidKeypairError extends AeError { }
+
+export class UnAvailableAccountError extends AeError {
+  constructor (address: string) {
+    super(`Account for ${address} not available`)
```
davidyuk, line 24:
```suggestion
    super(`Account for ${address} not available`)
    this.name = 'UnAvailableAccountError';
```
to change "Error:" to "UnAvailableAccountError:" when error is converted to string
based on https://stackoverflow.com/a/32750746

subhod-i:
@mradkov  @davidyuk Noted down the suggestions. I will update them accordingly.

src/utils/error.ts
subhod-i: Works the same but the existing one is marked invalid by the linter. https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/restrict-template-expressions.md

src/utils/error.ts
subhod-i: Removing it.

src/utils/error.ts
subhod-i: Done.

src/account/memory.js
subhod-i: and generate a public key from the private key?

davidyuk: yep, anyway, currently we are doing it for validation

PR #1345, opened 2021-12-01
feat(error): introduce error types

subhod-i:
Fix #1194 

Added custom error classes for the account, ae, chain, contract, typings, index, and node files.
Remaining: channel, node-pool, tx, and the utils.

Incorporated all the changes requested from the stale PR #1339

PR #1351, opened 2021-12-06
Introduce error class in Channel and TX source-code files

### LINKED ISSUE #1194: Introduce different error types

subhod-i:
Fix #1194 
source branch: channel and tx errors
target branch: Overall SDK error classes. Current has an open PR against develop branch https://github.com/aeternity/aepp-sdk-js/pull/1345

### DIFF
--- a/src/channel/handlers.js
+++ b/src/channel/handlers.js
@@ -29,7 +29,7 @@ import {
 } from './internal'
 import { unpackTx, buildTx } from '../tx/builder'
 import {
-  IlleagalArgumentError,
+  IllegalArgumentError,
   InsufficientBalanceError,
   ChannelConnectionError,
   UnexpectedChannelMessageError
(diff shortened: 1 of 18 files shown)


src/channel/handlers.js
```diff
@@ -28,6 +28,12 @@ import {
   fsmId
 } from './internal'
 import { unpackTx, buildTx } from '../tx/builder'
+import {
+  IlleagalArgumentError,
```
marc0olo, line 32:
```suggestion
  IllegalArgumentError,
```

we already had this error type in the SDK, right? we should fix the spelling if we do a major release anyway 😅

subhod-i: @marc0olo We had no such error in the SDK. No no, It's unacceptable irrespective of release. I will fix the spelling.

subhod-i: Fixed it.

test/unit/tx.js
```diff
@@ -100,32 +106,32 @@ describe('Tx', function () {
   })
 
   it('classify: invalid hash', () => {
-    expect(() => classify('aaaaa')).to.throw('Not a valid hash')
+    expect(() => classify('aaaaa')).to.throw(InvalidHashError, 'Not a valid hash')
```
marc0olo, line 109: where do we use `classify` internally? in another PR we renamed it because I thought it's only used to determine pointer keys or is this another function? see https://github.com/aeternity/aepp-sdk-js/pull/1334#discussion_r763200657

@davidyuk maybe you can also comment this 😅

PR #1353, opened 2021-12-06
test(contract aci): cover Chain.ttl

### LINKED ISSUE #1159: Passing Chain.ttl type
As per ACI, 
image

Chain.ttl is a type, and the compiler doesn't complain having that in an `entrypoint` contrary to higher order functions.
```
    entrypoint time_remaining(value : Chain.ttl) : Chain.ttl = 
        value 
```
Now I don't know whether this is intended to be possible and if yes, how is one supposed to call that in the JS SDK? Tried several variations but the type checker is always complaining.

davidyuk:
closes #1159


marc0olo (approved):
lgtm

mradkov (approved):

PR #1356, opened 2021-12-07
Release 10.0.0

davidyuk:
depends on #1355 #1334 #1183


mradkov:
> depends on #1355 #1334 #1183

All of those are now merged. Can you rebase and mark for review :)

mradkov (approved):

marc0olo (approved):
looks good to me.

I am wondering regarding the pointer keys if we should restrict the amount of pointers to max. `32` now, too. but we can also add this later as it wouldn't be a breaking change anyway.

PR #1359, opened 2021-12-07
Strict decode and related fixes

davidyuk:


src/tx/builder/schema.js
```diff
@@ -752,8 +752,8 @@ const CHANNEL_FORCE_PROGRESS_TX = [
   TX_FIELD('fromId', FIELD_TYPES.id, 'ak'),
   TX_FIELD('payload', FIELD_TYPES.binary, 'tx'),
   TX_FIELD('round', FIELD_TYPES.int),
-  TX_FIELD('update', FIELD_TYPES.binary),
-  TX_FIELD('stateHash', FIELD_TYPES.binary),
+  TX_FIELD('update', FIELD_TYPES.binary, 'cb'),
```
davidyuk, line 755: Actually, I'm not sure about this one, can't find a clear definition in both:
https://github.com/aeternity/protocol/blob/master/serializations.md#channel-solo-force-progress-transaction and
https://github.com/aeternity/protocol/blob/master/node/api/api_encoding.md#encoding-scheme-for-api-identifiers-and-byte-arrays

src/tx/builder/schema.js
```diff
@@ -752,8 +752,8 @@ const CHANNEL_FORCE_PROGRESS_TX = [
...
+  TX_FIELD('stateHash', FIELD_TYPES.binary, 'st'),
```
davidyuk, line 756: Ensured that this is correct by requesting it from middleware 😃 
https://mainnet.aeternity.io/mdw/txs/backward?type=channel_force_progress&limit=10&page=1

src/tx/builder/helpers.js
```diff
@@ -106,6 +104,22 @@ export function commitmentHash (name, salt = createSalt()) {
...
 
+// based on https://github.com/aeternity/protocol/blob/master/node/api/api_encoding.md
+const base64Types = ['ba', 'cb', 'or', 'ov', 'pi', 'ss', 'cs', 'ck', 'cv', 'st', 'tx']
+const base58Types = ['ak', 'bf', 'bs', 'bx', 'ch', 'cm', 'ct', 'kh', 'mh', 'nm', 'ok', 'oq', 'pp', 'sg', 'th']
+// TODO: add all types with a fixed length
+const typesLength = {
+  ak: 32,
+  ct: 32,
+  ok: 32
+}
```
subhod-i, line 115: Could be moved to `schema.js` file in the same directory. schema file holds all the constants as of now.

subhod-i (approved):

thepiwo (approved):
good as far as I understand

src/tx/builder/helpers.js
davidyuk: I would like to keep it local because it is not intended to be used somewhere else. Also putting it to `schema.js` will make it a part of public api, making it more difficult to change it.

PR #1361, opened 2021-12-08
Implement remote contract events support

### LINKED ISSUE #1344: handle events in remote contracts
as we recently noticed there is currently no way for the SDK to handle events of remote contracts.

thanks to @thepiwo we now at least ignore unknown events after this PR:
- https://github.com/aeternity/aepp-sdk-js/pull/1337/files

we should investigate this and find a way how to handle that situation. for the dex implementation we want to be able to write tests that depend on events emitted from a remote contract.

IMO we should at least be able to handle this if we know the ACI of all contracts involved in the interaction.

davidyuk:
depends on https://github.com/aeternity/aepp-calldata-js/pull/112
depends on #1395 
closes #1344 except for events with the same name in different contracts


marc0olo:
that's awesome! =)

do we have any idea to handle same names?

I didn't check in detail yet how it works with this PR. do we rely on any additional http request or is it fully covered with calldata-lib in the sdk?

davidyuk:
> do we have any idea to handle same names?

nope

> do we rely on any additional http request or is it fully covered with calldata-lib in the sdk?

the events before were decoded by sdk, without requests to http compiler or so, in this PR I've removed our poor implementation and switched to calldata

test/integration/contract-aci.js
```diff
@@ -520,7 +543,7 @@ describe('Contract instance', function () {
 
       it('Invalid arguments length', async () => {
         await expect(testContract.methods.datTypeGFn())
-          .to.be.rejectedWith('Non matching number of arguments. Got 0 but expected from 1 to 1')
+          .to.be.rejectedWith('Non matching number of arguments. datTypeGFn expects between 1 and 1 number of arguments but got 0')
```
thepiwo, line 546: weird error message

thepiwo (approved):
pretty good if I understand correctly

davidyuk:
> do we have any idea to handle same names?

I've added `contractAddressToName` option to resolve this

marc0olo (approved):
lgtm

test/integration/contract-aci.js
marc0olo: @davidyuk does this message come from calldata-lib ? just wondering

thepiwo (approved):

Issue #1309, opened 2021-10-25
improve error message for `out of gas` errors


marc0olo:
there was some discussion around that during the past days. I just tested the debug logs in AEstudio and got following error log in the console when a tx ran out of gas:

![image](https://user-images.githubusercontent.com/8124114/138660414-ff6ade62-53e9-4264-bd36-393f5de7af9d.png)

https://github.com/aeternity/aepp-sdk-js/blob/6dc63cbc87a21301b7d7934aece4a13a31be58d6/src/ae/contract.js#L60

the developer should actually get an information that the tx ran out of gas instead of just receiving "Error: Invocation failed".

davidyuk:
The node doesn't seem to provide that info, `getTransactionInfoByHash` returns object like:
```json
{
  "callerId": "ak_2WV5aAk3SCpVzm13q6mvvybbKaLj1YYyjFB93AEvWvNXxKLg5w",
  "callerNonce": 5,
  "contractId": "ct_55eqP7c5CKiVBfGPoqssa5xYpvdtp7uWLgCHirqhNNvMavZyH",
  "gasPrice": 1000000000,
  "gasUsed": 10,
  "height": 764,
  "log": [],
  "returnType": "error",
  "returnValue": "cb_Xfbg4g=="
}
```
`cb_Xfbg4g==` using compiler decodes to
```json
{
  "error": [
    ""
  ]
}
```
We can assume that a call run out of gas if `gasUsed` is equal to gas limit. Also, we can call txDryRun for the same tx:
```json
{
  "callerId": "ak_2Q5WkttXYzLBtt7sBivaRWUabBNc1niHKoYtB7m6vgkgfJMmVs",
  "callerNonce": 5,
  "contractId": "ct_q5R3jcSimsztC7pvJvXNEHm5qemsUFZhE53TydEaSM9sfRiCd",
  "gasPrice": 1000000000,
  "gasUsed": 10,
  "height": 1128,
  "log": [],
  "returnType": "error",
  "returnValue": "cb_T3V0IG9mIGdhc4mjSdQ="
}
```
`cb_T3V0IG9mIGdhc4mjSdQ=` using compiler decodes to
```json
{
  "error": [
    "Out of gas"
  ]
}
```
The best to solve it I think to make the node report "Out of gas" in on-chain call

marc0olo:
I opened an issue in the node repo: https://github.com/aeternity/aeternity/issues/3775

Issue #1286, opened 2021-10-11
Optional automatic gas consumption estimation 


nikita-fuchs:
Current issue: We used to have a high gas limit for stateful calls. Now we lowered it. The consequence of this is that many calls done in `aestudio` will fail for out of gas, inconveniently requiring people to poke around with the gas amount and misleading them into thinking their code might be the fault.

Possible fix: We provide a tx option that 

- estimates gas consumption 
- by dry-running the stateful call
- with a maximum gas limit of a microblock 

this value shall be then taken as gas amount for the statefull call (including contract deployment). If there is a gas amount provided though, _that_ value should be taken and override the auto-estimation option for quick testing (open for other suggestions here).

marc0olo:
when speaking about gas estimation behavior for stateful and non-stateful contract calls then I see multiple options here:

## General
- allow to set `gas_limit` manually (should already be the case ;))

## Non-stateful, static calls
- allow to choose between debug and protected dry-run endpoint
- for debug endpoint use a very high limit as default (e.g. `1_500_000` or even more)
- for protected dry-run endpoint use the max-default possible as default

## Stateful calls
- do **NOT** call dry-run for gas-estimation by default
- use always the default `25000` as `gas_limit` in case no `gas_limit` is explicitely set
- provide an option to automatically calculate expected gas consumption via dry-run giving an additional offset (e.g. x1.5)

what do the others think here?

nikita-fuchs:
> @nikita-fuchs is AEstudio using this workaround right now?

No it's using the default dryrun still.

marc0olo:
> No it's using the default dryrun still.

maybe you can change that along with the SDK update ;-) ... would be a quick win here as long as we haven't decided how to proceed here

kenodressel:
> do NOT call dry-run for gas-estimation by default
I disagree here because this means there will be a lot of errors where tx will run out of gas and users will not really know what to do. Obviously this comes with the performance pain and we might want to issue a warning based on this but I think by default the SDK should be easy to use and not necessarily the fastest thing in the world.

marc0olo:
for automatic gas consumption estimation we should also handle errors accordingly. see discussion here:
- https://github.com/aeternity/aeternity/issues/3775#issuecomment-966262039

davidyuk:
My proposal:
- to drop the default gas limit in sdk (it sounds like "640K is enough for anyone")
[...]

PR #1367, opened 2021-12-10
feat(aci): use dry-run to estimate gas and get rich errors

davidyuk:
closes https://github.com/aeternity/aepp-sdk-js/issues/1286
closes https://github.com/aeternity/aepp-sdk-js/issues/1309
closes https://github.com/aeternity/aeternity/issues/3775

### DIFF
--- a/src/tx/builder/schema.js
+++ b/src/tx/builder/schema.js
@@ -22,7 +22,6 @@ export const RESPONSE_TTL = { type: 'delta', value: 10 }
 // # CONTRACT
 export const DEPOSIT = 0
 export const AMOUNT = 0
-export const GAS_ABOVE_AVERAGE = 25000 // TODO: don't use this
 export const GAS_MAX = 1600000 - 21000
 export const MIN_GAS_PRICE = 1e9
 export const MAX_AUTH_FUN_GAS = 50000
--- a/src/tx/builder/schema.js
+++ b/src/tx/builder/schema.js
@@ -22,7 +22,8 @@ export const RESPONSE_TTL = { type: 'delta', value: 10 }
 // # CONTRACT
 export const DEPOSIT = 0
 export const AMOUNT = 0
-export const GAS = 25000
+export const GAS_ABOVE_AVERAGE = 25000 // TODO: don't use this
+export const GAS_MAX = 1600000 - 21000
 export const MIN_GAS_PRICE = 1e9
 export const MAX_AUTH_FUN_GAS = 50000
 export const DRY_RUN_ACCOUNT = { pub: 'ak_11111111111111111111111111111111273Yts', amount: '100000000000000000000000000000000000' }
--- a/docs/transaction-options.md
+++ b/docs/transaction-options.md
@@ -28,7 +28,7 @@ These options are common and can be provided to every tx-type:
     - 2 different strategies to use in order to determine the next nonce, See option `strategy` to learn more.
 - `strategy` (default: `max`)
     - The strategy to obtain next nonce for an account via node API
-    - If set to `max`, then the greatest nonce seen in the account or currently in the transaction pool is incremented with 1 and returned. 
+    - If set to `max`, then the greatest nonce seen in the account or currently in the transaction pool is incremented with 1 and returned.
     If the strategy is set to `continuity`, then transactions in the mempool are checked if there are gaps - missing nonces that prevent transactions with greater nonces to get included
 - `ttl` (default: `0`)
     - Should be set if you want the transaction to be only valid until a certain block height is reached.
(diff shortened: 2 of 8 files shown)


marc0olo (commented):
thanks @davidyuk! this is without an offset for the gas estimation now, right? I still think we should introduce an offset to the dry-run estimation as it just ensures that the tx gets included in case a little bit more gas is required. it won't have any effect on the

src/tx/builder/schema.js
```diff
@@ -22,7 +22,8 @@ export const RESPONSE_TTL = { type: 'delta', value: 10 }
 // # CONTRACT
 export const DEPOSIT = 0
 export const AMOUNT = 0
-export const GAS = 25000
+export const GAS_ABOVE_AVERAGE = 25000 // TODO: don't use this
+export const GAS_MAX = 1600000 - 21000
```
marc0olo, line 26: why do you want to introduce this again? this was the default in the past. we should avoid this being used

src/tx/builder/schema.js
```diff
@@ -22,7 +22,8 @@ export const RESPONSE_TTL = { type: 'delta', value: 10 }
 // # CONTRACT
 export const DEPOSIT = 0
 export const AMOUNT = 0
-export const GAS = 25000
+export const GAS_ABOVE_AVERAGE = 25000 // TODO: don't use this
```
marc0olo, line 25: why do we need this?

davidyuk: `gaAttachTx` still depends on it, I would like to deal with it later

davidyuk:
> I still think we should introduce an offset to the dry-run estimation

completely agree, here is it:
https://github.com/aeternity/aepp-sdk-js/pull/1367/files#diff-c6ada54b9fae5ec59ecdacbeaf2c48a0c3c1d9c81fc76c8c55fd52dcb9e1e958R160-R165

src/tx/builder/schema.js
marc0olo: mhh, I think I get your point. you need this for transactions that do not support dry-runs, correct? I just consulted with @hanssv and got confirmed that the only tx-types that aren't supported by the dry-run endpoint are:
- `GaMetaTx`
- `PayingForTx`

for these transactions we can use the default `25000` in my opinion. I would also remove the `TODO:` wording in the comment. or do you see any other solution? I currently don't 😅

src/tx/builder/schema.js
marc0olo: when I read the docs correctly we don't need to provide this param for a dry-run. according to the docs it will be automatically set to `1_000_000`, see https://github.com/aeternity/aeternity/blob/master/apps/aehttp/priv/oas3.yaml#L580

can you check that? as we are using protected dry-run by default anyway advanced devs should know how to deal with gas estimation anyway.

src/tx/builder/schema.js
subhod-i, line 25: `GAS_ABOVE_AVERAGE` is confusing. Do you mean this is the default Gas limit used for methods not supported by dry-run?

thepiwo: We should maybe make gas a required parameter for those then, as otherwise it will be unclear why executions fail?

src/tx/builder/schema.js
davidyuk: > we don't need to provide this param for a dry-run

> Since DryRunCallReq object do not have a mandatory gas field, if not set a default value of 1000000 is being used instead.

Gas can be omitted only if we pass transaction as an object instead of a tx-encoded string.

https://github.com/aeternity/aeternity/blob/6f0113e00766d0b675aab0327292b9d270a1f052/apps/aehttp/priv/oas3.yaml#L3245-L3266

We didn't use this feature before, and for me looks more robust to pass as an encoded string because we also checking that the string is encoded correctly.

src/tx/builder/schema.js
davidyuk: The same estimation as for contract deployment seems to work fine, so I'm removing this constant.

davidyuk: > Do you mean this is the default Gas limit used for methods not supported by dry-run?

Yep. In particular, this is the gas value that is above the average gas of contract calls/deployments. I was using it in `createGeneralizeAccount` but seems that it is not needed there.

marc0olo: is this a problem with the force pushs or how can I see what changed in regards to our comments? I am really struggling to understand what we are doing now.

I understand you removed `GAS_ABOVE_AVERAGE` as you can reuse another estimation.

sorry to ask / comment again:
- `GaAttachTx` can estimated via dry-run according to documentation and Hans => do we do that now? I don't understand why we need to use the logic for contract deployment here 🤔 
- how are we dealing now for `GaMetaTx` and `PayingForTx` ? what amount of `GAS` do we use there per default? should we open another ticket for this?

src/tx/builder/schema.js
marc0olo: ok I understand. but why do we use  `GAS_MAX` again? (if it is even a correct value for the max gas)

where do we need this and why?

davidyuk: I'm using `GAS_MAX` in dry-run calls (static calls of a contract method) if the user didn't provide any. The gas field can't be omitted because the transaction is tx-encoded at that moment. SDK can't pass any other gas value, because it can be estimated only by this call.

marc0olo: ok, got it

marc0olo (approved):
looks goot to me

thepiwo (approved):

PR #1362, opened 2021-12-08
test(integration): revert exit flag to speedup tests

davidyuk:
channel tests doesn't stop by themself without `exit` flag, I have accidentally removed it in https://github.com/aeternity/aepp-sdk-js/commit/bf15c1e07980c34b0197dd65b118ba39c3ee3fee#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L22-R22

PR #1368, opened 2021-12-11
Remove deprecated contract methods

davidyuk:
depends on #1362
depends on #1363


thepiwo:
whats the goal of this and changes it introduces?

davidyuk:
I've rebased it on the top of develop to cleanup diff.

Basically, I propose to remove contract-related methods that do the same as can be done using contracts generated by the `getContractInstance` method. `getContractInstance` is a preferred method, so removing this would make API clearer.

thepiwo (approved):

PR #1371, opened 2021-12-17
feat(package): build es version compatible with node

davidyuk:
depends on #1369 (needed for cli)

So, after this would be possible to do:
```js
import { Crypto } from '@aeternity/aepp-sdk/es/index.mjs'

console.log(Crypto.generateKeyPair())
```
and execute it using plain node:
```
$ node test.mjs
{
  publicKey: 'ak_jnZsCzXTKdGDkdvSpExo46GwuWGv84CkzcvEtd6XdV98gLjd8',
  secretKey: 'cb8556c8ff170c0075c3a8bdfa0343c83782e1ccbca0e7842a0caaee92c10fce61266723868d53aab664039e4088d72df242a1510f2bab2e9c67dbeb768ac1ad'
}
$ node --version                                            
v16.13.0
```

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -4,7 +4,6 @@
     "@babel/preset-typescript"
   ],
   "plugins": [
-    ["@babel/plugin-transform-runtime", { "corejs": 3 }],
-    "ramda"
+    ["@babel/plugin-transform-runtime", { "corejs": 3 }]
   ]
 }
--- a/babel.esm.config.js
+++ b/babel.esm.config.js
@@ -1,5 +1,7 @@
 const config = require('./babel.config.json')
 
 config.presets[0] = ['@babel/preset-env', { modules: false }]
+config.plugins.push(['add-import-extension', { extension: 'mjs' }])
+config.plugins.push('transform-default-named-imports')
 
 module.exports = config
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,8 @@
         "@typescript-eslint/eslint-plugin": "^5.3.1",
         "@typescript-eslint/parser": "^5.3.1",
         "babel-loader": "^8.2.3",
+        "babel-plugin-add-import-extension": "^1.6.0",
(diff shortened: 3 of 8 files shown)


subhod-i (approved):

.github/workflows/main.yml
```diff
@@ -4,19 +4,13 @@ jobs:
...
-      - if: github.ref != 'refs/heads/develop'
-        uses: actions/checkout@v2
-        with:
-          ref: develop
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 100
-      - if: github.ref != 'refs/heads/develop'
-        run: git branch develop origin/develop
+          fetch-depth: 50
```
thepiwo, line 9: why do we need fetch-depth at all?

thepiwo: ah I see, to lint the last 50 commits

src/chain/index.js
```diff
@@ -37,7 +37,27 @@ import { required } from '@stamp/required'
  * @return {Object} Chain instance
  */
 const Chain = stampit({
-  deepProps: { Ae: { defaults: { waitMined: true, verify: true } } }
+  deepProps: {
+    Ae: {
+      defaults: {
```
thepiwo, line 42: why is this needed here?

src/chain/index.js
```diff
@@ -37,7 +37,27 @@ import { required } from '@stamp/required'
...
+        waitMined: true,
+        verify: true,
+        _expectedMineRate: 180000,
+        _microBlockCycle: 3000,
+        _maxPollInterval: 5000
+      }
+    }
+  },
+  methods: {
+    _getPollInterval (type) {
```
thepiwo, line 52: why is this needed here?

thepiwo (approved):

src/chain/index.js
davidyuk: so, these all are changes from #1369, I've rebased this PR on the top of develop and they are not visible in the diff anymore

PR #1374, opened 2021-12-20
fix(contract events): don't throw error if events emitted by remote

davidyuk:
depends on #1362

### DIFF
--- a/test/integration/contract-aci.js
+++ b/test/integration/contract-aci.js
@@ -395,6 +395,17 @@ describe('Contract instance', function () {
         }
       }])
     })
+
+    it('calls a contract that emits events with no defined events', async () => {
+      const contract = await sdk.getContractInstance({
+        source:
+          'contract FooContract =\n' +
+          '  entrypoint emitEvents(f: bool) = ()',
+        contractAddress: remoteContract.deployInfo.address
+      })
+      const result = await contract.methods.emitEvents(false, { omitUnknown: true })
+      expect(result.decodedEvents).to.be.eql([])
+    })
   })
 
   describe('Arguments Validation and Casting', function () {


test/integration/contract-aci.js
```diff
@@ -301,6 +301,20 @@ describe('Contract instance', function () {
 
       expect(testContract.decodeEvents(log)).to.be.eql(eventResult.decodedEvents)
     })
+
+    it('calls a contract without events that calls another contract emitting events', async () => {
+      const contract = await sdk.getContractInstance({
+        source:
+          'contract interface StateContract =\n' +
+          '  stateful entrypoint emitEvents: () => unit\n' +
```
subhod-i, line 309: @davidyuk not sure if this method is emitting any events? What does `unit` stands for? Looked into the Sophia documentation, Did not find anything on it.

subhod-i: Got it. L308 contract is just an interface for instance creation. Thank you.

test/integration/contract-aci.js
```diff
@@ -301,6 +301,20 @@ describe('Contract instance', function () {
...
+          '\n' +
+          'contract FooContract =\n' +
+          '  entrypoint foo(s: StateContract) = s.emitEvents()'
+      })
+      await contract.deploy()
+      const result = await contract.methods.foo(testContract.deployInfo.address)
+      expect(result.decodedEvents).to.be.eql([])
```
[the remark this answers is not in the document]
davidyuk: only in `result.result.log`, they are omitted in `decodedEvents` because they are not defined in the contract interface

thepiwo (approved):

davidyuk:
original bug was fixed in #1361, so merging only test

PR #954, opened 2020-03-30
feat(Chain): Add option `allowUnsynced` for `poll` method

nduchak:

PR #1376, opened 2021-12-24
refactor: deprecate tx and getTxInfo methods

davidyuk:
depends on #1375

### DIFF
--- a/src/chain/node.js
+++ b/src/chain/node.js
@@ -79,9 +79,9 @@ async function sendTransaction (tx, options = {}) {
 
 async function waitForTxConfirm (txHash, options = { confirm: 3 }) {
   options.confirm = options.confirm === true ? 3 : options.confirm
-  const { blockHeight } = await this.tx(txHash)
+  const { blockHeight } = await this.api.getTransactionByHash(txHash)
   const height = await this.awaitHeight(blockHeight + options.confirm, options)
-  const { blockHeight: newBlockHeight } = await this.tx(txHash)
+  const { blockHeight: newBlockHeight } = await this.api.getTransactionByHash(txHash)
   switch (newBlockHeight) {
     case -1:
       throw new TxNotInChainError(txHash)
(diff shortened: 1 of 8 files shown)


src/chain/node.js
```diff
@@ -141,18 +144,24 @@ async function topBlock () {
   return this.api.getTopHeader()
 }
 
-async function poll (th, { blocks = 10, interval = 500, allowUnsynced = false } = {}) {
```
davidyuk, line 144: `allowUnsynced` was implemented in https://github.com/aeternity/aepp-sdk-js/pull/954 and used only in FireEditor (added in https://github.com/aeternity/fire-editor/commit/85ed8a39be2024b28eb873d7261ae53d90f95a74)

I can't find any explanation of reasoning in both repositories, but as I got the `allowUnsynced` flag makes SDK check `txInfo.callInfo.height`, but from the first view it should be the same as `tx.blockHeight`. It was made to improve performance, but I don't think that it has significant benefits in comparison to using calldata library and other improvements that we introduced recently. SDK tests work relatively fast without `allowUnsynced`, switching this flag to true doesn't improve performance. So I'm proposing to drop it. cc: @nikita-fuchs

src/chain/node.js
```diff
@@ -141,18 +144,24 @@ async function topBlock () {
...
+    ...status ? [`status: ${status}`] : []
+  ].join(', '))
 }
 
+/**
+ * @deprecated use `sdk.api.getTransactionInfoByHash` instead
+ */
 async function getTxInfo (hash) {
   const result = await this.api.getTransactionInfoByHash(hash)
   return result.callInfo || result
```
davidyuk, line 167: `getTransactionInfoByHash` also has `gaInfo` and `txInfo` fields, merging this would make it overcomplicated, I'm proposing to use the plain endpoint of node instead.

src/chain/node.js
```diff
@@ -103,6 +103,9 @@ async function getBalance (address, { height, hash, format = AE_AMOUNT_FORMATS.A
   return formatAmount(balance, { targetDenomination: format }).toString()
 }
 
+/**
+ * @deprecated use `sdk.api.getTransactionByHash/getTransactionInfoByHash` instead
+ */
 async function tx (hash, info = true) {
   const tx = await this.api.getTransactionByHash(hash)
   if (['ContractCreateTx', 'ContractCallTx', 'ChannelForceProgressTx'].includes(tx.tx.type) && info && tx.blockHeight !== -1) {
```
davidyuk, line 111: For these transactions types, it would make two requests to the node even if txInfo is not needed. Also, we have additional transaction types that have non-empty transaction info, seems irrational to add them here, so proposing to use plain endpoint instead.

src/chain/node.js
thepiwo: sounds good to me

thepiwo (approved):

PR #1380, opened 2021-12-26
Making a sequence of requests in efficient way

### LINKED ISSUE #963: Add ability to prepare and send batch of transactions
**Is your feature request related to a problem? Please describe.**
We need to come up with some implementation which will help developers to work with a batch of transaction

davidyuk:
depends on #1362 
closes #963
reported issue https://github.com/aeternity/aeternity/issues/3803

### DIFF
--- a/src/utils/swagger.js
+++ b/src/utils/swagger.js
@@ -115,6 +115,7 @@ export default async (
     .flat()
     .reduce((acc, n) => ({ ...acc, [n.operationId]: n }), {})
 
+  const requestQueues = {}
   const api = mapObject(combinedApi, ([opId, handler]) => [
     opId.slice(0, 1).toLowerCase() + snakeToPascal(opId.slice(1)),
     async (...args) => {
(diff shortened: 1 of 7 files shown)


thepiwo (approved):
cool!

src/utils/swagger.js
```diff
@@ -62,9 +62,25 @@ export default async (
...
+    const pendingGetRequests = {}
     return SwaggerClient({
       url,
       spec,
+      requestInterceptor: request => {
+        if (request.method !== 'GET') return
+        return {
+          ...request,
+          userFetch: async (url, request) => {
+            const key = JSON.stringify({ ...request, url })
```
subhod-i, line 74: @davidyuk We could hash the key to make it memory-friendly and unique.

davidyuk: I thought about something like this, but the request seems to be a small amount of data to optimize its storage. It is overkill to use third-party hashing packages, and integrated crypto would force to use promises. Keeping this easy-to-read seems more important than this optimization.

Possible collisions may add the ability for a malicious actor to replace the response content by requesting a specially-constructed URL before another request, but this is very theoretical.

subhod-i: Since it's a key-based structure rather than an array, replay attacks are anyway eliminated. If the request is small then we have no issues.

davidyuk: > Since it's a key-based structure rather than an array, replay attacks are anyway eliminated

Not sure about this. Let's assume we have two names pointing to different addresses and a function that will send tokens to address resolved by the name (this is a bad idea, but let's assume we have it). The attacker can't change that name, but he can request other name details right before that function is called.

The correct name details requested by https://mainnet.aeternity.io/v3/names/superhero.chain, the hash is X. The attacker can add meaningless data to name/URL so theoretically he can find a URL like https://mainnet.aeternity.io/v3/names/malicious.chain?t=Y that has the same X hash. The last thing to do is to request that URL right before calling a function that will spend tokens. The Swagger will request only the first URL and propagate the malicious name details to both requests.

PR #1384, opened 2021-12-29
Refactor oracle

davidyuk:
depends on #1362


thepiwo (commented):
seems to overlap a lot with #1385 whats the plan there?

davidyuk:
Yes, this is a subset of #1385 that is not blocked by the node issue. Merging this as #1385 approved.

PR #1385, opened 2021-12-29
perf(oracle): use from parameter of GetOracleQueriesByPubkey

### LINKED ISSUE #1055: Optimise of pollForQueries by using `from` parameter of GetOracleQueriesByPubkey

davidyuk:
depends on #1384 
blocked by https://github.com/aeternity/aeternity/issues/3805
closes #1055

### DIFF
--- a/src/ae/oracle.js
+++ b/src/ae/oracle.js
@@ -72,12 +72,11 @@ function pollForQueries (
   onQuery,
   { interval = this._getPollInterval('microblock') } = {}
 ) {
-  const knownQueryIds = new Set()
+  let from
   const checkNewQueries = async () => {
-    const queries = ((await this.api.getOracleQueriesByPubkey(oracleId)).oracleQueries || [])
-      .filter(({ id }) => !knownQueryIds.has(id))
-    queries.forEach(({ id }) => knownQueryIds.add(id))
-    if (queries.length) onQuery(queries)
+    const { oracleQueries } = await this.api.getOracleQueriesByPubkey(oracleId)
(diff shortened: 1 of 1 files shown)


src/ae/oracle.js
```diff
@@ -68,17 +67,22 @@ async function getOracleObject (oracleId) {
...
+    if (oracleQueries.length) onQuery(oracleQueries)
   }
 
-  checkNewQueries()
-  const intervalId = setInterval(checkNewQueries, interval)
-  return () => clearInterval(intervalId)
+  let stopped
+  (async () => {
+    while (!stopped) { // eslint-disable-line no-unmodified-loop-condition
+      // TODO: allow to handle this error somehow
```
thepiwo, line 80: so this is still work in progress?

thepiwo (approved):

src/ae/oracle.js
davidyuk: Nope, we need some api to handle poll exceptions, but it is out of the scope of this PR.

Probably this PR would be closed because it looks like I misunderstood the meaning of `from` parameter (see the related issue).

marc0olo:
@thepiwo what's that about? @davidyuk mentioned that we should probably close that one. how should we proceed?

subhod-i:
This PR depends on https://github.com/aeternity/aeternity/issues/3805 . Not sure the SDK will work without the core changes.

marc0olo:
let's wait for feedback here. I don't think this issue is urgent

davidyuk:
`from` is not working as I expected, node team consider not feasible to change it, so closing this one

PR #1388, opened 2021-12-29
fix(tx builder): reject more than 32 pointers

### LINKED ISSUE #1358: AENS: limit the amount of pointers to max. 32
as follow up of https://github.com/aeternity/aepp-sdk-js/issues/1240 - I think we missed to address this there.

davidyuk:
depends on #1362 
closes #1358


thepiwo (approved):
looks good to me

PR #1407, opened 2022-02-09
Remove node stream polyfill

davidyuk:
I'm planning to rename `TxBuildHelper.decode/encode` to something more general, but it is not ready yet.

The need to switch from `bs58check` to `bs58` is explained in https://github.com/aeternity/aepp-calldata-js/pull/122.

[bip39](https://www.npmjs.com/package/bip39) package can be configured to support different languages and this is not up to SDK to decide which of them should be included. Also, it depends on a set of unnecessary crypto libraries that is not easy to get rid of: https://github.com/bitcoinjs/bip39/issues/170. It has a quite simple blockchain-agnostic interface. Our `@aeternity/bip39` didn't get updates in 3 years and probably shouldn't be used. So, I'm proposing to make the wallet developer use `bip39` package by himself and remove it from sdk.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "SDK"
   ],
   "dependencies": {
-    "@aeternity/aepp-calldata": "^1.1.0",
+    "@aeternity/aepp-calldata": "github:aeternity/aepp-calldata-js#a623a44a259cc53e99b1e358ecc7b3f88edcb45c",
     "@aeternity/argon2-browser": "^0.1.1",
     "@aeternity/json-bigint": "^0.3.1",
     "@aeternity/uuid": "^0.0.1",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -34,9 +34,8 @@
     "SDK"
   ],
   "dependencies": {
-    "@aeternity/aepp-calldata": "^1.1.0",
+    "@aeternity/aepp-calldata": "github:aeternity/aepp-calldata-js#a623a44a259cc53e99b1e358ecc7b3f88edcb45c",
```
thepiwo, line 37: will there be a release with these changes included?

thepiwo (approved):

package.json
davidyuk: Yep, I think so

PR #1414, opened 2022-02-11
Use class-based fields in TxBuilder

### LINKED ISSUE #1266: nameFee is requested in calculateMinFee function
In wallet we were using v1 of `nameClaim` schema and after updating to 0.8.2 sdk v1 has been removed, so we switched to v2. We are using that schema to calculate min fee through `calculateMinFee` function. It appeared that calling of this function with `vsn: 2` led to this error in sdk:
image

After some research I've found that this `formatAmount` is calling with `nameFee: undefined` because we are not passing this parameter to function.
Is this a bug or we really should provide some default `nameFee` (which sdk actually has in `import { getMinimumNameFee } from '@aeternity/aepp-sdk/es/tx/builder/helpers';`) for the function which is calculating the fee?

davidyuk:
closes #1266

### DIFF
--- a/src/ae/aens.js
+++ b/src/ae/aens.js
@@ -28,7 +28,7 @@
 
 import { salt } from '../utils/crypto'
 import {
-  commitmentHash, ensureNameValid, getMinimumNameFee, isAuctionName, encode, produceNameId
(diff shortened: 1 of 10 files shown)


thepiwo (approved):
looks cool

PR #1418, opened 2022-02-22
fix(delegate signature): don't encode address depending on onAccount

davidyuk:
probably fixes a bug in #1274 
found issue: https://github.com/aeternity/aepp-sdk-js/issues/1419


marc0olo:
> probably fixes a bug in #1274

might this be the reason for this to fail? https://github.com/aeternity/aepp-sophia-examples/blob/master/test/oracleDelegation.js#L66

davidyuk:
> might this be the reason for this to fail?

Yep, see https://github.com/aeternity/aepp-sophia-examples/pull/69

subhod-i (approved):

PR #1427, opened 2022-03-16
Release v11.0.0

subhod-i:
- [x] Add migration guide
- [x] Update changelog


thepiwo (commented):
looks really good so far, I will have another look though

marc0olo:
same here. need to have a more comprehensive look into changelog and migration guide tomorrow

src/ae/oracle.js
```diff
@@ -76,9 +80,15 @@ function pollForQueries (oracleId, onQuery, { interval = 5000 } = {}) {
     if (queries.length) onQuery(queries)
   }
 
-  checkNewQueries()
-  const intervalId = setInterval(checkNewQueries, interval)
-  return () => clearInterval(intervalId)
+  let stopped
+  (async () => {
+    while (!stopped) { // eslint-disable-line no-unmodified-loop-condition
+      // TODO: allow to handle this error somehow
```
thepiwo, line 86: still an open todo?

src/chain/index.js
```diff
@@ -37,13 +37,32 @@ import { required } from '@stamp/required'
...
+    }
+  },
+  methods: {
+    _getPollInterval (type) {
+      const base = {
+        block: this.Ae.defaults._expectedMineRate,
+        microblock: this.Ae.defaults._microBlockCycle
+      }[type]
+      if (!base) throw new Error(`Unknown poll type: ${type}`)
+      return Math.min(base / 3, this.Ae.defaults._maxPollInterval)
```
thepiwo, line 58: Shouldn't the methods and config above either moved to a config then applied here or be somehow the defaults? I don't understand why this is specified here.

thepiwo (commented):
- cool new docs, awesome compatibility table
- overall looks really good
- great cleanups
- nice new error handling
- my same comments apply for the aeSdk renaming (@marc0olo)

Is tree-shaking still possible (and needed?) if so is there a documentation for this?

Haven't had the time to try it out in detail yet, but code looks really good, minor comments remain.

Awesome work on this!

subhod-i:
> Is tree-shaking still possible (and needed?) if so is there a documentation for this?
>  Do we have tree-shaking configuration set? I see a note on this here in the contribution document https://github.com/aeternity/aepp-sdk-js/tree/develop/docs/contrib#principles

IMO,  `tree-shaking` using webpack is not the ideal way. It depends on marking the files as pure or side-effects. Not sure if the benefit is worth more than the effort considering our codebase size.

src/ae/oracle.js
subhod-i: @davidyuk  is now displaying the error on the console. TODO is open but I don't think it needs immediate attention.

subhod-i: best way is throwing a custom error type.

thepiwo: ok :)

src/chain/index.js
subhod-i: these parameters are not configurable for the developers. `getPollInterval` method is used in case the option `interval` is not specified. 

e.g.

```js
function pollForQueries (
  oracleId,
  onQuery,
  { interval = this._getPollInterval('microblock') } = {}
) {
  const knownQueryIds = new Set()
  ....
  }
  ```
  
  Value of the defaults could be moved to `schema.js` file.

marc0olo: `schema.js` is the wrong place for this IMO as it's mainly (or even exclusively) used for tx related config values.

subhod-i: Okay. Makes sense. I remember having the same discussing with Denis.

marc0olo (approved):
from my side it looks good now. @subhod-i I removed the codecov stuff and updated the docs and readme again. I identified some small issues and removed outdated and irrelevant badges

if we really want and need codecov we should revisit after release again. this topic just annoys me currently.

PR #1438, opened 2022-03-24
skip payable check for method `init`

### LINKED ISSUE #1436: SDK requires `payable` for `init` function
seems like we introduced a bad behavior 😞 

in aepp-sophia-examples we provide AE to the `init` function while deploying a contract and this used to work with the previous version.

now with v11.0.0 the SDK complains:
```
NotPayableFunctionError: You try to pay "You try to pay "100000" to function "init" which is not payable. Only payable function can accept coins" to function "undefined" which is not payable. Only payable function can accept tokens
```

independently of that this error message seems to be strange, too.

when I add (for test-purposes) the `payable` declaration to the `init` function the compiler complains as follows:
```
Error: aci error:
type_error:14:3: The init function (at line 14, column 3) cannot be payable.
You don't need the 'payable' annotation to be able to attach
funds to the create contract transaction.
```

subhod-i:
Closes #1436

### DIFF
--- a/src/contract/aci.js
+++ b/src/contract/aci.js
@@ -235,7 +235,7 @@ export default async function getContractInstance ({
     if (!fn) throw new MissingFunctionNameError()
     if (fn === 'init' && !opt.callStatic) throw new InvalidMethodInvocationError('"init" can be called only via dryRun')
     if (!contractId && fn !== 'init') throw new InvalidMethodInvocationError('You need to deploy contract before calling!')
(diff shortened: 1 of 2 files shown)


subhod-i:
@marc0olo @thepiwo 
I haven't bumped up the package version.
Testing this branch in `aeproject` and `aepp-sophia-examples`

subhod-i:
Creating a separate PR with all the fixes

PR #1439, opened 2022-03-25
Various hotfixes 🤷‍♀️

### LINKED ISSUE #1436: SDK requires `payable` for `init` function

### LINKED ISSUE #1437: oracle delegation signature
again for aepp-sophia-examples. the tests used to work and now fail for some reason after updating to v11.0.0:

```sh
1) OracleDelegation
       Interact with the contract
         should register an oracle:
     NodeInvocationError: Invocation failed: "Error in oracle_register: bad_signature"
      at l (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:94920)
      at Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97516)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Object.c._estimateGas (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:95715)
      at async Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97636)
      at async Context.<anonymous> (test/oracleDelegation.js:47:13)

  2) OracleDelegation
       Interact with the contract
         should extend an oracle:
     NodeInvocationError: Invocation failed: "Error in oracle_extend: bad_signature"
      at l (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:94920)
      at Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97516)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Object.c._estimateGas (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:95715)
      at async Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97636)
      at async Context.<anonymous> (test/oracleDelegation.js:53:13)

  3) OracleDelegation
       Interact with the contract
         should respond to a specific query:
     NodeInvocationError: Invocation failed: "oracle not valid"
      at l (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:94920)
      at Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97516)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Object.c._estimateGas (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:95715)
      at async Object.c.call (node_modules/@aeternity/aepp-sdk/dist/aepp-sdk.js:1:97636)
      at async Context.<anonymous> (test/oracleDelegation.js:58:39)
 ```

### PR DESCRIPTION
**Pull request by subhod-i**
skip payable check for method init: closes #1436 
don't include the address for oracle respond signature: closes #1437 

edit: Partially closes #1436

### DIFF
--- a/src/ae/contract.js
+++ b/src/ae/contract.js
@@ -106,7 +106,10 @@ async function createAensDelegationSignature ({ contractId, name }, opt = {}) {
  * const respondSig = await contract.createOracleDelegationSignature(params, queryId)
  */
 async function createOracleDelegationSignature ({ contractId, queryId }, opt = {}) {
-  return this.delegateSignatureCommon([...queryId ? [queryId] : [], contractId], opt)
+  return this.delegateSignatureCommon(
+    [...queryId ? [queryId] : [await this.address(opt)], contractId],
+    opt
+  )
 }
 
 /**
--- a/src/contract/aci.js
+++ b/src/contract/aci.js
@@ -235,7 +235,7 @@ export default async function getContractInstance ({
     if (!fn) throw new MissingFunctionNameError()
     if (fn === 'init' && !opt.callStatic) throw new InvalidMethodInvocationError('"init" can be called only via dryRun')
     if (!contractId && fn !== 'init') throw new InvalidMethodInvocationError('You need to deploy contract before calling!')
-    if (opt.amount > 0 && fnACI.payable === false) throw new NotPayableFunctionError(`You try to pay "${opt.amount}" to function "${fn}" which is not payable. Only payable function can accept coins`)
+    if (fn !== 'init' && opt.amount > 0 && fnACI.payable === false) throw new NotPayableFunctionError(`You try to pay "${opt.amount}" to function "${fn}" which is not payable. Only payable function can accept coins`)
 
     const callerId = await this.address(opt).catch(error => {
       if (opt.callStatic) return DRY_RUN_ACCOUNT.pub
--- a/test/integration/contract-aci.js
+++ b/test/integration/contract-aci.js
@@ -305,20 +305,7 @@ describe('Contract instance', function () {
   })
 
   describe('Transaction options', async () => {
-    it('deploys with no options', async () => {
-      const testContract = await aeSdk.getContractInstance({
-        source: testContractSource, filesystem
-      })
-      const deployInfo = await testContract.deploy(['test', 1, 'hahahaha'])
-      expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
-      expect(deployInfo.txData.tx.gas).to.be.above(0)
-      expect(deployInfo.txData.tx.amount).to.be.equal(0)
-      expect(deployInfo.txData.gasUsed).to.be.above(0)
-      expect(testContract.bytecode).to.satisfy(b => b.startsWith('cb_'))
-      testContractAddress = deployInfo.address
-    })
-
-    it('deploys with options', async () => {
+    it('deploys', async () => {
       const testContract = await aeSdk.getContractInstance({
         source: testContractSource, filesystem
       })
@@ -332,7 +319,6 @@ describe('Contract instance', function () {
         strategy: 'max'
       })
       expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
-      console.log(deployInfo.txData)
       expect(deployInfo.txData.tx.gas).to.be.equal(15000)
       expect(deployInfo.txData.tx.amount).to.be.equal(42)
       expect(deployInfo.txData.gasUsed).to.be.equal(209)
--- a/test/integration/contract-aci.js
+++ b/test/integration/contract-aci.js
@@ -304,6 +304,45 @@ describe('Contract instance', function () {
     result.callerId.should.be.equal(onAccount)
   })
 
+  describe('Transaction options', async () => {
+    it('deploys with no options', async () => {
+      const testContract = await aeSdk.getContractInstance({
+        source: testContractSource, filesystem
+      })
+      const deployInfo = await testContract.deploy(['test', 1, 'hahahaha'])
+      expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
+      expect(deployInfo.txData.tx.gas).to.be.above(0)
+      expect(deployInfo.txData.tx.amount).to.be.equal(0)
+      expect(deployInfo.txData.gasUsed).to.be.above(0)
+      expect(testContract.bytecode).to.satisfy(b => b.startsWith('cb_'))
+      testContractAddress = deployInfo.address
+    })
+
+    it('deploys with options', async () => {
+      const testContract = await aeSdk.getContractInstance({
+        source: testContractSource, filesystem
+      })
+      const deployInfo = await testContract.deploy(['test', 1, 'hahahaha'], {
+        amount: 42,
+        denomination: 'aettos',
+        gas: 15000,
+        deposit: 0,
+        ttl: 0,
+        gasPrice: '1e9',
+        strategy: 'max'
+      })
+      expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
+      console.log(deployInfo.txData)
+      expect(deployInfo.txData.tx.gas).to.be.equal(15000)
+      expect(deployInfo.txData.tx.amount).to.be.equal(42)
+      expect(deployInfo.txData.gasUsed).to.be.equal(209)
+      expect(deployInfo.txData.gasPrice).to.be.equal(1000000000)
+      expect(deployInfo.txData.tx.deposit).to.be.equal(0)
+      expect(testContract.bytecode).to.satisfy(b => b.startsWith('cb_'))
+      testContractAddress = deployInfo.address
+    })
+  })
+
   describe('Gas', () => {
     let contract
 
--- a/test/integration/contract.js
+++ b/test/integration/contract.js
@@ -413,7 +413,7 @@ describe('Contract', function () {
     })
 
     it('registers', async () => {
-      delegationSignature = await aeSdk.delegateSignatureCommon([address, contractId])
+      delegationSignature = await aeSdk.createOracleDelegationSignature({ contractId })
       const oracleRegister = await contract.methods.signedRegisterOracle(
         address, delegationSignature, queryFee, ttl
       )
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-mkdocs==1.2.3
+mkdocs==1.2.4
 mkdocs-simple-hooks==0.1.3
 mkdocs-material==7.1.9
 mike==1.0.1

### CODE REVIEW

File: test/integration/contract.js [hunk 1]
```diff
@@ -413,7 +413,7 @@ describe('Contract', function () {
     })
 
     it('registers', async () => {
-      delegationSignature = await aeSdk.delegateSignatureCommon([address, contractId])
+      delegationSignature = await aeSdk.createOracleDelegationSignature({ contractId })
```
marc0olo, line 416: mhh, do we introduce another breaking change with this?

marc0olo (commented):
does sophia examples run now with these fixes? I will check separately before approval. not sure if I find the time today

test/integration/contract.js
subhod-i: Not a breaking change. `delegateSignatureCommon` or `createOracleDelegationSignature` any method can be used  and syntax remained same.

src/contract/aci.js
```diff
@@ -235,7 +235,7 @@ export default async function getContractInstance ({
     if (!fn) throw new MissingFunctionNameError()
     if (fn === 'init' && !opt.callStatic) throw new InvalidMethodInvocationError('"init" can be called only via dryRun')
     if (!contractId && fn !== 'init') throw new InvalidMethodInvocationError('You need to deploy contract before calling!')
-    if (opt.amount > 0 && fnACI.payable === false) throw new NotPayableFunctionError(`You try to pay "${opt.amount}" to function "${fn}" which is not payable. Only payable function can accept coins`)
+    if (fn !== 'init' && opt.amount > 0 && fnACI.payable === false) throw new NotPayableFunctionError(`You try to pay "${opt.amount}" to function "${fn}" which is not payable. Only payable function can accept coins`)
```
davidyuk, line 238: would work to change `{ payable: false }` to `{ payable: true }` in `getFunctionACI` instead?

subhod-i: Function call sometimes never toches that line when the function ACI is already present in the instance. if statement in here.  https://github.com/aeternity/aepp-sdk-js/blob/5fe6cac41718fd7dd8645a1cf5034bdb654067a6/src/contract/aci.js#L215

davidyuk: we can do it later after implementing https://github.com/aeternity/aesophia/issues/373

test/integration/contract.js
davidyuk: I would switch to `delegateSignatureCommon` at the end because wrapper functions just places it's arguments into array, they are not needed.

test/integration/contract-aci.js
```diff
@@ -304,6 +304,45 @@ describe('Contract instance', function () {
...
+        amount: 42,
+        denomination: 'aettos',
+        gas: 15000,
+        deposit: 0,
+        ttl: 0,
+        gasPrice: '1e9',
+        strategy: 'max'
+      })
+      expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
+      console.log(deployInfo.txData)
```
davidyuk, line 335:
```suggestion
```

test/integration/contract-aci.js
```diff
@@ -304,6 +304,45 @@ describe('Contract instance', function () {
...
+      const deployInfo = await testContract.deploy(['test', 1, 'hahahaha'])
+      expect(deployInfo.address).to.satisfy(b => b.startsWith('ct_'))
+      expect(deployInfo.txData.tx.gas).to.be.above(0)
+      expect(deployInfo.txData.tx.amount).to.be.equal(0)
+      expect(deployInfo.txData.gasUsed).to.be.above(0)
+      expect(testContract.bytecode).to.satisfy(b => b.startsWith('cb_'))
+      testContractAddress = deployInfo.address
+    })
+
+    it('deploys with options', async () => {
```
davidyuk, line 321: I would update the "deploys" test instead: https://github.com/aeternity/aepp-sdk-js/blob/2cc3b9137c1d1c88dcce5eb6875188d198690ce8/test/integration/contract-aci.js#L184-L192

test/integration/contract-aci.js
```diff
@@ -304,6 +304,45 @@ describe('Contract instance', function () {
     result.callerId.should.be.equal(onAccount)
   })
 
+  describe('Transaction options', async () => {
+    it('deploys with no options', async () => {
+      const testContract = await aeSdk.getContractInstance({
+        source: testContractSource, filesystem
+      })
+      const deployInfo = await testContract.deploy(['test', 1, 'hahahaha'])
```
davidyuk, line 312: Deployment with no options used multiple times in SDK tests, I don't think that it is needed to be tested explicitly.

marc0olo:
@subhod-i @davidyuk FYI => I provided this PR because there is a problem right now with mkdocs `1.2.3` which causes a failed docs pipeline. this should be included in hotfix IMO

test/integration/contract-aci.js
subhod-i: Sure

test/integration/contract-aci.js
subhod-i: Sure

test/integration/contract-aci.js
subhod-i: Sneaky! Will delete it.

src/contract/aci.js
subhod-i: Do we need a new issue here regarding `init` is  `payble` or not?

marc0olo: from current discussion in that issue above  `init` will never be `payable`. we should discuss that in the issue @davidyuk opened in aesophia repo.

I guess current solution here is good as is probably. need your statements about that

test/integration/contract.js
subhod-i: someone reading the client code it would be more pleasing if we keep `createOracleDelegationSignature`. It's more readable. having `delegateSignatureCommon` is very confusing.

src/contract/aci.js
subhod-i: Yeah, I read the thread in aesophia issue. `init` is never callable which means we can't estimate gas for deployment as of now which is wrong, we should be able to estimate gas for deployment calls. 

My conclusion is: Let's keep the changes. Ideally, the compiler throws an error anyway. the source of truth is with the compiler/node. That means the following client-side check actually doesn't make sense. It's just saving a request from the compiler. The only problem I see in here with the fix we did is **SDK now allows sending amount during deployment call even if the contract is not marked as payable**. Well, the node will return an error anyway😀
https://github.com/aeternity/aepp-sdk-js/blob/5fe6cac41718fd7dd8645a1cf5034bdb654067a6/src/contract/aci.js#L238

marc0olo: > we can't estimate gas for deployment as of now which is wrong, we should be able to estimate gas for deployment calls

who said that? actually it is possible to estimate gas afaik.

> The only problem I see in here with the fix we did is **SDK now allows sending amount during deployment call even if the contract is not marked as payable**. Well, the node will return an error anyway😀

I agree, but if we don't do it people can't provide AE even to `payable` contracts, right? ... oh man 😵 

> Well, the node will return an error anyway😀

it won't !!!! this is the reason why @davidyuk opened the issue in aesophia repository I guess. we actually can send AE to a non-payable contract during deployment with the `ContractCreateTx`.

mhhh 🤔 

what do you think, @kenodressel @thepiwo ?

subhod-i: > who said that? actually it is possible to estimate gas afaik.  

I mean without the fix in this PR. 

> I agree, but if we don't do it people can't provide AE even to payable contracts, right? ... oh man. 

Yes. They still can provided `gas` option.

> it won't !!!! this is the reason why @davidyuk opened the issue in aesophia repository I guess. we actually can send AE to a non-payable contract during deployment with the ContractCreateTx

Wow! I hope this is the case only for `contractCreateTx`. Imagine compiler depending on client call ACI for a contract method. This unblocks a new fear of attack. A contract method ACI can be marked payable at the client-side and so on..

marc0olo: > Wow! I hope this is the case only for `contractCreateTx`. Imagine compiler depending on client call ACI for a contract method. This unblocks a new fear of attack. A contract method ACI can be marked payable at the client-side and so on..

yes, only for deployment (or `init`, but we shouldn't say `init` is used because it somehow has no meaning except setting the initial state)

marc0olo (approved):
I think we clarified everything here, right? approved from my side.

@davidyuk any comment from you before we merge?

subhod-i:
@marc0olo Changing the target branch to develop.

src/contract/aci.js
davidyuk: The current version is fine 👌

src/ae/contract.js
```diff
@@ -106,7 +106,10 @@ async function createAensDelegationSignature ({ contractId, name }, opt = {}) {
  * const respondSig = await contract.createOracleDelegationSignature(params, queryId)
  */
 async function createOracleDelegationSignature ({ contractId, queryId }, opt = {}) {
-  return this.delegateSignatureCommon([...queryId ? [queryId] : [], contractId], opt)
+  return this.delegateSignatureCommon(
+    [...queryId ? [queryId] : [await this.address(opt)], contractId],
```
davidyuk, line 110:
```suggestion
    [queryId ? queryId : await this.address(opt), contractId],
```

test/integration/contract.js
davidyuk: "entities should not be multiplied beyond necessity", `createAensDelegationSignature`, `createOracleDelegationSignature` just signing an array of arguments in the right order. I think user can pass them by himself, just according to documentation, like:
```
aens: address, name, contractId
oracle: queryId, contractId
```
Also, I don't like that `createOracleDelegationSignature` actually can be used to create AENS delegation signature and probably even to sign an arbitrary transaction.

But let's keep it as it is right now.

davidyuk:
> Changing the target branch to develop

I would merge this to `master` firstly, and then realign `master` to `develop` after release.

test/integration/contract-aci.js
```diff
@@ -304,6 +304,31 @@ describe('Contract instance', function () {
     result.callerId.should.be.equal(onAccount)
   })
 
+  describe('Transaction options', async () => {
+    it('deploys', async () => {
```
davidyuk, line 308: we already have `deploys` test in this file, proposing to combine them

subhod-i:
Closing this in favor of https://github.com/aeternity/aepp-sdk-js/pull/1462

Issue #1421, opened 2022-03-01
Unable to build test aepp for production, found circular deps in SDK


quantum0hound:
When I serve an aepp to production, I receive such error in browser as: 
```
schema.js:475 Uncaught ReferenceError: Cannot access 'dr' before initialization
    at Object.52373 (schema.js:475:20)
    at n (bootstrap:19:32)
    at startup:4:87
    at Function.n.O (chunk loaded:25:12)
    at startup:5:43
    at app.2e26927d.js:1:114649
```
I've found that error is caused by WebPack bundling and found [https://github.com/webpack/webpack/issues/12724].
Then I've added circular-dependency-plugin to vue.config.js and received:
```
 ERROR  Failed to compile with 6 errors                                                                                                                                                1:57:44 PM

 error  

Circular dependency detected:
src/@aeternity/aepp-sdk/es/channel/handlers.mjs -> src/@aeternity/aepp-sdk/es/channel/internal.mjs -> src/@aeternity/aepp-sdk/es/channel/handlers.mjs

 error  

Circular dependency detected:
src/@aeternity/aepp-sdk/es/channel/internal.mjs -> src/@aeternity/aepp-sdk/es/channel/handlers.mjs -> src/@aeternity/aepp-sdk/es/channel/internal.mjs

 error  

Circular dependency detected:
src/@aeternity/aepp-sdk/es/tx/builder/field-types.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/helpers.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/schema.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/field-types.mjs

 error  

Circular dependency detected:
src/@aeternity/aepp-sdk/es/tx/builder/helpers.mjs -> src/@aeternity/aepp-sdk/es/utils/crypto.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/helpers.mjs

 error  

Circular dependency detected:
src/@aeternity/aepp-sdk/es/tx/builder/schema.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/field-types.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/helpers.mjs -> src/@aeternity/aepp-sdk/es/tx/builder/schema.mjs

 error  

Circular dependency detected:
[...]

PR #1442, opened 2022-03-29
refactor: avoid cycle dependencies

davidyuk:
Cycle deps topic was raised in #1421 it is reasonable to fix


src/tx/builder/constants.js
```diff
@@ -0,0 +1,100 @@
+import BigNumber from 'bignumber.js'
```
subhod-i, line 1: The schemas here are widely imported and used by the community developers. If we are moving them to a new file then we better add this to the migration guide. Super important!

subhod-i (approved):

src/tx/builder/constants.js
subhod-i: @davidyuk

davidyuk: it is not a breaking change because I have reexported them as it was before:
https://github.com/aeternity/aepp-sdk-js/pull/1442/files#diff-ad38e160e81f081497c7c5146c2f092580b6e1fb9989415eae179c477dfbfb84R13
also
https://github.com/aeternity/aepp-sdk-js/pull/1442/files#diff-e496a95ddc06df5bbd37ef86211584337612c734f283228563dddb48e7b406deR33
https://github.com/aeternity/aepp-sdk-js/pull/1442/files#diff-119538feced9a8a6925bd1421ca798809cda6c6823bbff34af904a036f4b0d6bR33

PR #1446, opened 2022-03-30
refactor: update instructions of Quick Start

janmichek:
As I went through examples and guides when I was learning about Aeternity I found some clutters.
My motivation here is to make developers onboard smoothly and leave no doubts in that process.
Nothing is written in stone, here are my suggestions.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -1,63 +1,63 @@
 # Quick Start
+In this example we will send 1 _AE_ coin from one account to another
 
 ## 1. Specify imports
 For the following snippets in the guide you need to specify multiple imports. Most imports like `Universal`, `MemoryAccount` & `Node` are [Stamps](https://stampit.js.org/essentials/what-is-a-stamp) that compose certain functionalities. Others utility functions like `generateKeyPair` also can be imported.
 
 ```js
-import {
+const {
   Universal,
   MemoryAccount,
   Node,
   AE_AMOUNT_FORMATS,
   generateKeyPair
-} from '@aeternity/aepp-sdk'
+} = require('@aeternity/aepp-sdk')
 ```
 
-## 2. Create a Keypair
+## 2. Create a Keypair for sender
 
 ```js
-  const keypair = generateKeyPair()
-  console.log(`Secret key: ${keypair.secretKey}`)
-  console.log(`Public key: ${keypair.publicKey}`)
+const keypair = generateKeyPair()
+console.log(`Secret key: ${keypair.secretKey}`)
+console.log(`Public key: ${keypair.publicKey}`)
 ```
 
 ## 3. Get some _AE_ using the Faucet
-To receive some _AE_ you can use the [Faucet](https://faucet.aepps.com/). Just add your publicKey, and you'll immediately get some test coins.
+To receive some _AE_ you can use the [Faucet](https://faucet.aepps.com/). Just paste sender's publicKey, hit `Top UP` and you'll immediately get some test coins.
 
 ## 4. Interact with the æternity blockchain
 This example shows:
 
 - how to create an instance of the SDK using the `Universal` [Stamp](https://stampit.js.org/essentials/what-is-a-stamp)
-- how to get the current block height
-- how to spend 1 AE from the account the SDK instance was initialized with to some other AE address
+- how to spend (send) 1 AE from the account the SDK instance was initialized with to some other AE address
 
 ```js
-const NODE_URL = 'https://testnet.aeternity.io';
-const COMPILER_URL = 'https://compiler.aepps.com'; // required for contract interactions (compiling and creating calldata)
-const account = MemoryAccount({
-  // replace <YOUR_SECRET> and <YOUR_PUBLIC_KEY> with the values from step 2
-  keypair: { secretKey: '<YOUR_SECRET>', publicKey: '<YOUR_PUBLIC_KEY>' }
-})
+const NODE_URL = 'https://testnet.aeternity.io'
+const COMPILER_URL = 'https://compiler.aepps.com' // required for contract interactions
+const senderAccount = MemoryAccount({
+  // replace <SENDER_SECRET> and <SENDER_PUBLIC_KEY> with the generated keypair from step 2
+  keypair: { secretKey: '<SENDER_SECRET>', publicKey: '<SENDER_PUBLIC_KEY>' }
+});
 
 (async function () {
   const node = await Node({ url: NODE_URL })
   const aeSdk = await Universal({
-     compilerUrl: COMPILER_URL,
-     nodes: [ { name: 'testnet', instance: node } ],
-     accounts: [ account ]
+    compilerUrl: COMPILER_URL,
+    nodes: [{ name: 'testnet', instance: node }],
+    accounts: [senderAccount]
   })
 
-  const height = await aeSdk.height() // get top keyblock height
-  console.log('Current Keyblock Height:', height)
-
   // spend one AE
-  await aeSdk.spend(1, '<RECIPIENT_PUBLIC_KEY>', { // replace <RECIPIENT_PUBLIC_KEY>
-      denomination: AE_AMOUNT_FORMATS.AE
+  await aeSdk.spend(1, '<RECIPIENT_PUBLIC_KEY>', {
+    // replace <RECIPIENT_PUBLIC_KEY>, Ideally you use public key from Superhero Wallet you have created before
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -1,19 +1,20 @@
 # Quick Start
+In this example we will send 1 _AE_ coin from one account to another
 
 ## 1. Specify imports
 For the following snippets in the guide you need to specify multiple imports. Most imports like `Universal`, `MemoryAccount` & `Node` are [Stamps](https://stampit.js.org/essentials/what-is-a-stamp) that compose certain functionalities. Others like `AmountFormatter` & `Crypto` are util modules with typical function exports.
 
 ```js
-import {
+const {
```
janmichek, line 8: I assume the developers will create `somePlainJsFile.js` and start coding

Personally, I would also use modern import, but this older one works out of the box.

docs/quick-start.md
```diff
@@ -1,19 +1,20 @@
...
   MemoryAccount,
   Node,
   AmountFormatter,
   Crypto
-} from '@aeternity/aepp-sdk'
+} = require('@aeternity/aepp-sdk')
 ```
 
-## 2. Create a Keypair
+## 2. Create a Keypair for sender
```
janmichek, line 17: I wanted to explain there is a Sender and Recipient.
`<RECIPIENT_PUBLIC_KEY>` was set before, but there was no note about Sender

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
-To receive some _AE_ you can use the [Faucet](https://faucet.aepps.com/). Just add your publicKey, and you'll immediately get some test coins.
+To receive some _AE_ you can use the [Faucet](https://faucet.aepps.com/). Just paste sender's publicKey, hit `Top UP` and you'll immediately get some test coins.
 
 ## 4. Interact with the æternity blockchain
 This example shows:
 
 - how to create an instance of the SDK using the `Universal` [Stamp](https://stampit.js.org/essentials/what-is-a-stamp)
-- how to get the current block height
-- how to spend 1 AE from the account the SDK instance was initialized with to some other AE address
+- how to spend (send) 1 AE from the account the SDK instance was initialized with to some other AE address
```
janmichek, line 32: I suggest to omit printing block height as this was covered in here
https://aeternity.com/aepp-sdk-js/v11.0.0/
Would focus on only one thing which is sending coin from between accounts

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
 
 ```js
 const NODE_URL = 'https://testnet.aeternity.io';
-const COMPILER_URL = 'https://compiler.aepps.com'; // required for contract interactions (compiling and creating calldata)
-const account = MemoryAccount({
-  // replace <YOUR_SECRET> and <YOUR_PUBLIC_KEY> with the values from step 2
-  keypair: { secretKey: '<YOUR_SECRET>', publicKey: '<YOUR_PUBLIC_KEY>' }
+const COMPILER_URL = 'https://compiler.aepps.com'; // required for contract interactions
```
janmichek, line 36: I guess the info was needed and caused unnecessary horizontal scroll
firefox_11ZB0mQ6AQ

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
+     accounts: [ senderAccount ]
   })
 
-  const height = await aeSdk.height() // get top keyblock height
-  console.log('Current Keyblock Height:', height)
-
   // spend one AE
-  await aeSdk.spend(1, '<RECIPIENT_PUBLIC_KEY>', { // replace <RECIPIENT_PUBLIC_KEY>
+  await aeSdk.spend(1, '<RECIPIENT_PUBLIC_KEY>', {
+      // replace <RECIPIENT_PUBLIC_KEY>, Ideally you use public key from Superhero Wallet you have created before
```
janmichek, line 52: Here I wanted to point developers to set up Superhero.

Otherwise we would need to guide developers to generate another (recipient) account.
Make sense to me as this is the very first example.

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
 ```
 
 Note:
 
+- You may remove code from Step 2 as this serves only for one-time creation
 - By default the `spend` function expects the amount to be spent in `aettos` (the smallest possible unit)
-- Following the example snippet you would specify `AE` as denomination
\ No newline at end of file
+- Following the example snippet you would specify `AE` as denomination
+- See [Testnet Explorer](https://explorer.testnet.aeternity.io/) and track your transactions
```
janmichek, line 63: I think its a nice way to introduce Explorer in here and overall observe the result of example given

docs/quick-start.md
janmichek: Do we want to explain it in plain JS context or in Vue context?

subhod-i: We can't specific to one javascript platform. This is good, plain javascript!

subhod-i:
Everything looks good except one on which I have made a comment.

marc0olo:
> Everything looks good except one on which I have made a comment.

what should be changed?

@janmichek can you apply the change? :-)

docs/quick-start.md
janmichek: So this - older ES5 import will work out of the box with plain js

janmichek:
The change I made (using older import) will work

subhod-i:
> what should be changed?

I have added the comments.

janmichek:
> > what should be changed?
> 
> I have added the comments.

Yes, thanks for that. I think I have replied to your comment here
https://github.com/aeternity/aepp-sdk-js/pull/1446#discussion_r842608488

All should be solved and can be merged if you don't have any other comments @subhod-i

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
 const NODE_URL = 'https://testnet.aeternity.io';
-const COMPILER_URL = 'https://compiler.aepps.com'; // required for contract interactions (compiling and creating calldata)
-const account = MemoryAccount({
-  // replace <YOUR_SECRET> and <YOUR_PUBLIC_KEY> with the values from step 2
-  keypair: { secretKey: '<YOUR_SECRET>', publicKey: '<YOUR_PUBLIC_KEY>' }
+const COMPILER_URL = 'https://compiler.aepps.com'; // required for contract interactions
+const senderAccount = MemoryAccount({
+  // replace <SENDER_SECRET> and <SENDER_PUBLIC_KEY> with the generated keyppair from step 2
+  keypair: { secretKey: '<SENDER_SECRET>', publicKey: '<SENDER_PUBLIC_KEY>' }
 })
```
davidyuk, line 40: Does it works without semicolon here?

docs/quick-start.md
davidyuk:
```suggestion
    // replace <RECIPIENT_PUBLIC_KEY>, Ideally you use public key from Superhero Wallet you have created before
```
the indentation should be 2 spaces as I see

docs/quick-start.md
```diff
@@ -22,42 +23,41 @@ import {
...
 
 (async function () {
   const node = await Node({ url: NODE_URL })
   const aeSdk = await Universal({
      compilerUrl: COMPILER_URL,
```
davidyuk, line 45: This block uses 3 spaces 🤪

docs/quick-start.md
janmichek: Yes, that was my point.
In the modern type of import, there are some `if`s and ` requires a bit more experience` as you mentioned.
For that reason, I would suggest the older style of import as this is straightforward.
But feel free to suggest (describe) the way how to guide our users to use the modern way. 
I am open to it, but I preferred fewer steps in the guide.

docs/quick-start.md
janmichek: Interesting :D was like this even before, but I manage it to align it to 2 spaces

docs/quick-start.md
janmichek: I think so. Removed

docs/quick-start.md
janmichek: fixed

docs/quick-start.md
davidyuk: I mean it actually needed at the end of statement before a function defined and called at the same time:
```js
console.log('test')
(() => {})()
```
would fail with
> TypeError: console.log(...) is not a function

I've added this while rebasing

davidyuk (approved):

PR #1449, opened 2022-03-30
refactor: update instructions to wallet connection tutorial

janmichek:
Introducing updated instructions for wallet connection.
Now its in nice copy-pasteble form and cleared from unnecessary code.
Maybe I am not familiar with everything so please suggest changes

### DIFF
--- a/docs/guides/connect-aepp-to-wallet.md
+++ b/docs/guides/connect-aepp-to-wallet.md
@@ -22,7 +22,7 @@ export default {
     publicAddress: '',
     balance: 0,
     node: null,
-    accounts: null
+    connectedAccounts: null
   }),
 }
 ```
@@ -85,7 +85,7 @@ Append method for wallet connection
 ```js
 async connectToWallet(wallet) {
   await this.aeSdk.connectToWallet(await wallet.getConnection())
-  this.accounts = await this.aeSdk.subscribeAddress('subscribe', 'connected')
+  this.connectedAccounts = await this.aeSdk.subscribeAddress('subscribe', 'connected')
   this.publicAddress = await this.aeSdk.address()
   this.balance = await this.aeSdk.getBalance(this.publicAddress)
   this.node = await this.aeSdk.getNodeInfo()
--- a/docs/guides/connect-aepp-to-wallet.md
+++ b/docs/guides/connect-aepp-to-wallet.md
@@ -87,7 +87,7 @@ async connectToWallet(wallet) {
   await this.aeSdk.connectToWallet(await wallet.getConnection())
   this.connectedAccounts = await this.aeSdk.subscribeAddress('subscribe', 'connected')
   this.address = await this.aeSdk.address()
-  this.balance = await this.aeSdk.getBalance(this.address)
+  this.balance = await this.aeSdk.balance(this.address).catch(() => '0')
   this.nodeInfo = await this.aeSdk.getNodeInfo()
 }
 ```
--- a/docs/guides/connect-aepp-to-wallet.md
+++ b/docs/guides/connect-aepp-to-wallet.md
@@ -19,7 +19,7 @@ const COMPILER_URL = 'https://compiler.aepps.com';
 export default {
   data: () => ({
     aeSdk: null,
-    publicAddress: '',
+    address: '',
     balance: 0,
     nodeInfo: null,
     connectedAccounts: null
@@ -43,8 +43,8 @@ async created () {
       this.nodeInfo = await this.aeSdk.getNodeInfo()
     },
     onAddressChange: async () => {
-      this.publicAddress = await this.aeSdk.address()
-      this.balance = await this.aeSdk.balance(this.publicAddress)
+      this.address = await this.aeSdk.address()
+      this.balance = await this.aeSdk.balance(this.address)
     },
     onDisconnect: () => {
       // you may want to reset state here
@@ -86,8 +86,8 @@ Append method for wallet connection
 async connectToWallet(wallet) {
   await this.aeSdk.connectToWallet(await wallet.getConnection())
   this.connectedAccounts = await this.aeSdk.subscribeAddress('subscribe', 'connected')
-  this.publicAddress = await this.aeSdk.address()
-  this.balance = await this.aeSdk.getBalance(this.publicAddress)
+  this.address = await this.aeSdk.address()
+  this.balance = await this.aeSdk.getBalance(this.address)
   this.nodeInfo = await this.aeSdk.getNodeInfo()
 }
 ```
--- a/docs/guides/connect-aepp-to-wallet.md
+++ b/docs/guides/connect-aepp-to-wallet.md
@@ -64,7 +64,7 @@ methods: {
       if (confirm(`Do you want to connect to wallet ${newWallet.name}`)) {
         detector.stopScan()
         // connect to the wallet, see step 4.
-        await this.connectToWallet(newWallet)
+        await this.connect(newWallet)
       }
     }
 
@@ -83,7 +83,7 @@ methods: {
 Append method for wallet connection
 
 ```js
-async connectToWallet(wallet) {
+async connect(wallet) {
   await this.aeSdk.connectToWallet(await wallet.getConnection())
   this.connectedAccounts = await this.aeSdk.subscribeAddress('subscribe', 'connected')
   this.address = await this.aeSdk.address()
@@ -97,7 +97,7 @@ async connectToWallet(wallet) {
 Aepp can request the wallet to share its connected node URLs if any to interact with the chain.
 
 ```js
-async connectToWallet (wallet) {
+async connect (wallet) {
     await this.aeSdk.connectToWallet(await wallet.getConnection(), { connectNode: true, name: 'wallet-node', select: true })
 }
 ```
--- a/docs/guides/connect-aepp-to-wallet.md
+++ b/docs/guides/connect-aepp-to-wallet.md
@@ -1,90 +1,109 @@
 # Connect an æpp to a wallet
 
-## Introduction
+This guide describes the 4 steps that are necessary to connect your application to a wallet using the RPC API.
 
-This guide describes the 3 steps that are necessary to connect your application to a wallet using the RPC API.
+## Prerequisites
 
-Note:
+- Install [Superhero Wallet extension](https://wallet.superhero.com/) for simplicity of example.
+You can build your own wallet in the next example
 
-- The steps below are snippets taken from the full implementation of the [Simple æpp](https://github.com/aeternity/aepp-sdk-js/tree/master/examples/browser/aepp)
-
-## 1. Initialize the `RpcAepp` Stamp
+## 1. Specify imports and constants and state
 
 ```js
-import { RpcAepp, WalletDetector, BrowserWindowMessageConnection } from '@aeternity/aepp-sdk'
+import { RpcAepp, WalletDetector, BrowserWindowMessageConnection, Node } from '@aeternity/aepp-sdk'
 
+const TESTNET_NODE_URL = 'https://testnet.aeternity.io'
+const MAINNET_NODE_URL = 'https://mainnet.aeternity.io'
+const COMPILER_URL = 'https://compiler.aepps.com';
+
+export default {
+  data: () => ({
+    aeSdk: null,
+    publicAddress: '',
+    balance: 0,
+    node: null,
+    accounts: null
+  }),
+}
+```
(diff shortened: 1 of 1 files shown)


docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
 # Connect an æpp to a wallet
 
-## Introduction
+This guide describes the 4 steps that are necessary to connect your application to a wallet using the RPC API.
```
janmichek, line 3: some of the guides does not have this `Introduction` heading. Its kinda inconsistent. I think it can go without heading

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
 
-This guide describes the 3 steps that are necessary to connect your application to a wallet using the RPC API.
+## Prerequisites
 
-Note:
-
-- The steps below are snippets taken from the full implementation of the [Simple æpp](https://github.com/aeternity/aepp-sdk-js/tree/master/examples/browser/aepp)
+- Install [Superhero Wallet extension](https://wallet.superhero.com/) for simplicity of example.
```
janmichek, line 7: What you think about this added instruction. For the simplicity of example I would advise using Superhero

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
+You can build your own wallet in the next example
 
-## 1. Initialize the `RpcAepp` Stamp
+## 1. Specify imports and constants and state
 
 ```js
-import { RpcAepp, WalletDetector, BrowserWindowMessageConnection } from '@aeternity/aepp-sdk'
+import { RpcAepp, WalletDetector, BrowserWindowMessageConnection, Node } from '@aeternity/aepp-sdk'
```
janmichek, line 13: I added one more initialization step.
There was a lot of code missing and was not working before without going to https://github.com/aeternity/aepp-sdk-js/tree/master/examples/browser/aepp and start investigating

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
+    onNetworkChange: async ({ networkId }) => {
+      this.aeSdk.selectNode(networkId)
+      this.node = await this.aeSdk.getNodeInfo()
     },
-    onAddressChange:  async (addresses) => {
-      this.pub = await this.aeSdk.address()
-      this.balance = await this.aeSdk.balance(this.pub).catch(e => '0')
-      this.addressResponse = await this.aeSdk.address()
+    onAddressChange: async () => {
+      this.publicAddress = await this.aeSdk.address()
```
janmichek, line 46: `pub` -> `publicAddress` - I saw no good reason for shortening the name.

`this.addressResponse = await this.aeSdk.address()` was just duplicate

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
+      this.balance = await this.aeSdk.balance(this.publicAddress)
```
janmichek, line 47: Here I am not sure. For simplicity, I have ommitted sanitizing value.
The problem was that basic eslint configuration was complaining about unused variable `e`

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
     },
     onDisconnect: () => {
-      this.resetState()
+      // you may want to reset state here
```
janmichek, line 50: `this.resetState()` function was not really definet anywhere - or am I missing something?

I just put the comment for simplicity, but I may add reset function to reset state data

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
       alert('Disconnected')
     }
   })
-  this.height = await this.aeSdk.height()
-  // start looking for wallets, see step 2.
   this.scanForWallets()
-  // open iframe with Wallet if run in top window
-  if (window === window.parent) this.openReverseIframe()
```
janmichek, line 42: this function was not defined niether. I think its residue of this example https://github.com/aeternity/aepp-sdk-js/tree/master/examples/browser/aepp

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
-    if (confirm(`Do you want to connect to wallet ${newWallet.name}`)) {
-      this.detector.stopScan()
-      // connect to the wallet, see step 3.
-      await this.connectToWallet(newWallet)
+methods: {
+  async scanForWallets()  {
+    const handleWallets = async function ({ wallets, newWallet }) {
+      newWallet = newWallet || Object.values(wallets)[0]
+      if (confirm(`Do you want to connect to wallet ${newWallet.name}`)) {
+        detector.stopScan()
```
janmichek, line 65: detector does not have to be in state in here. This context is sufficient. Also residue of more complex example

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
     }
-  }
 
-  const scannerConnection = BrowserWindowMessageConnection({
-    connectionInfo: { id: 'spy' }
-  })
-  this.detector = WalletDetector({ connection: scannerConnection })
-  this.detector.scan(handleWallets.bind(this))
+    const scannerConnection = BrowserWindowMessageConnection({
+      connectionInfo: { id: 'spy' }
```
janmichek, line 72: Here would need some advise.
I would like to put comment in here but not really sure what this scanner is for. Maybe we can come up with better naming instead of comment?

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
 
+export default {
+  data: () => ({
+    aeSdk: null,
+    publicAddress: '',
+    balance: 0,
+    node: null,
+    connectedAccounts: null
+  }),
+}
```
janmichek, line 27: I am thinking of adding simple template just to print state variables.
There is no visual output of this example now.

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
-      { name: 'ae_uat', instance: await Node({ url: TEST_NET_NODE_URL }) },
-      { name: 'ae_mainnet', instance: await Node({ url: MAIN_NET_NODE_INTERNAL_URL }) }
+      { name: 'ae_uat', instance: await Node({ url: TESTNET_NODE_URL }) },
+      { name: 'ae_mainnet', instance: await Node({ url: MAINNET_NODE_URL }) }
     ],
     compilerUrl: COMPILER_URL,
-    onNetworkChange: async (params) => {
-      this.aeSdk.selectNode(params.networkId)
-      this.nodeInfoResponse = await this.aeSdk.getNodeInfo()
+    onNetworkChange: async ({ networkId }) => {
```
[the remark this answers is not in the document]
davidyuk: I would stay for `nodeInfo`, because just `node` can have a collision with instances of `Node` created in nodes array

docs/guides/connect-aepp-to-wallet.md
davidyuk: Is was defined before in aepp example, but I removed it while refactoring.

docs/guides/connect-aepp-to-wallet.md
davidyuk: The reverse iframe stuff is not needed in example, would be nice to make SDK handle it by itself someday 🙃

docs/guides/connect-aepp-to-wallet.md
davidyuk: I completely agree with no-shortening, but would name it just `address`, because we don't have any private addresses. Usually, I'm naming `ak_...` string an address, array of bytes a private/public key, but it is not aligned with naming in node.

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
+  this.publicAddress = await this.aeSdk.address()
+  this.balance = await this.aeSdk.getBalance(this.publicAddress)
+  this.node = await this.aeSdk.getNodeInfo()
 }
 ```
 
 ## 4. Use Wallet's Node for chain communication
 
-AEPP can request the wallet to share its connected node URLs if any to interact with the chain.
+Aepp can request the wallet to share its connected node URLs if any to interact with the chain.
```
davidyuk, line 97: Actually it is not a step at all, we need to decide should we use it by default or not cc: @subhod-i

docs/guides/connect-aepp-to-wallet.md
davidyuk: I think this guide was initially made vue-agnosting, not sure that we need to introduce specific stuff from these, just define a set of functions with comments.

docs/guides/connect-aepp-to-wallet.md
davidyuk:
```suggestion
      this.balance = await this.aeSdk.balance(this.publicAddress).catch(() => '0')
```
The node throws 404 for newly-created accounts, so the error handler would make it less confusing

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
-  this.pub = await this.aeSdk.address()
-  this.onAccount = this.pub
-  this.balance = await this.aeSdk.getBalance(this.pub)
-  this.walletName = this.aeSdk.rpcClient.info.name
-  this.addressResponse = await this.aeSdk.address()
-  this.heightResponse = await this.aeSdk.height()
-  this.nodeInfoResponse = await this.aeSdk.getNodeInfo()
+  this.publicAddress = await this.aeSdk.address()
+  this.balance = await this.aeSdk.getBalance(this.publicAddress)
+  this.node = await this.aeSdk.getNodeInfo()
```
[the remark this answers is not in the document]
davidyuk: `walletName` variable wasn't referenced here, so I'm fine to omit it

docs/guides/connect-aepp-to-wallet.md
davidyuk: `scannerConnection` is a stamp instance that polls an aepp with wallet info object until gets a reply 🤷‍♀️  I have a plan to simplify this process in future

davidyuk (approved):
In general looks good

docs/guides/connect-aepp-to-wallet.md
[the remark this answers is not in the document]
janmichek: Alright. For the sake of distinguishing can work as you suggested. Changed to `nodeInfo`

docs/guides/connect-aepp-to-wallet.md
janmichek: Good point, changed to `address`

docs/guides/connect-aepp-to-wallet.md
janmichek: Yes, thanks for the suggestion. Accepted

docs/guides/connect-aepp-to-wallet.md
janmichek: Ok, thanks for the explanation. I leave it without comment for now

docs/guides/connect-aepp-to-wallet.md
janmichek: leaving this with just a comment

docs/guides/connect-aepp-to-wallet.md
```diff
@@ -1,90 +1,109 @@
...
-      // connect to the wallet, see step 3.
-      await this.connectToWallet(newWallet)
+methods: {
+  async scanForWallets()  {
+    const handleWallets = async function ({ wallets, newWallet }) {
+      newWallet = newWallet || Object.values(wallets)[0]
+      if (confirm(`Do you want to connect to wallet ${newWallet.name}`)) {
+        detector.stopScan()
+        // connect to the wallet, see step 4.
+        await this.connect(newWallet)
```
janmichek, line 67: One more thing I have just fixed.
I found it out when I was working with RPC - and its propagated to most of the usages on GH

There are two functions named the same `connectToWallet`
One's source is in the example the other one is from SDK library.
I am finding this confusing, so I rename one just `connect`

Or suggest some other naming

davidyuk (approved):
👍

Issue #1318, opened 2021-11-03
Generate API reference documentation


subhod-i:
Auto-generate API reference documentation.

davidyuk:
implemented in https://github.com/aeternity/aepp-sdk-js/commit/6d5b7bdc6680a3bedaebce7b689b5566ad6536d3

PR #1451, opened 2022-03-30
docs: update instructions of contract guide

janmichek:
Just fixed some naming a bit, so there are no doubts which code we talking about.

I also have broader Idea in here. This page seems to me a bit more docs than practical example.

As a developer, I would like to see a straightforward working example. The guide here has a lot of aside explanation turnoffs and does not follow a straight line. So maybe it would deserve to create another section instead of wider explanation.
Let me know what you think.

The example I redone for myself ended up like this.

```
 const CONTRACT_SOURCE_CODE =
      `
      contract Increment =

    record state =
        { count: int }

    entrypoint init(start: int) =
        { count = start }

    stateful entrypoint increment(value: int) =
        put(state{ count = state.count + value })

    entrypoint get_count() =
        state.count
      `
    
    const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE_CODE.trim() })
    
    const deployTx = await contractInstance.deploy([1])
    console.log('deployTx', deployTx)
    
    const initialCountTx = await contractInstance.methods.get_count()
    console.log('initialCountTx', initialCountTx.decodedResult)
    
    const incrementTx = await contractInstance.methods.increment(3)
    console.log('incrementTx', incrementTx)
    
    const countAfterTx = await contractInstance.methods.get_count()
    console.log('countAfterTx', countAfterTx.decodedResult)
  },
```

### DIFF
--- a/docs/guides/contracts.md
+++ b/docs/guides/contracts.md
@@ -52,11 +52,11 @@ Note:
 
 ## 3. Initialize the contract instance
 
-### By sourcecode
+### By source code
 
 ```js
-const CONTRACT_SOURCE = ... // source code of the contract
-const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE })
+const sourceCode = ... // source code of the contract
+const contractInstance = await aeSdk.getContractInstance({ source: sourceCode })
 ```
 
 Note:
@@ -64,7 +64,7 @@ Note:
 - If your contract includes external dependencies which are not part of the [standard library](https://aeternity.com/aesophia/latest/sophia_stdlib) you should initialize the contract using:
   ```js
   const fileSystem = ... // key-value map with name of the include as key and source code of the include as value
-  const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE, fileSystem })
+  const contractInstance = await aeSdk.getContractInstance({ source: sourceCode, fileSystem })
   ```
 
 ### By ACI and bytecode
@@ -104,7 +104,7 @@ const contractInstance = await aeSdk.getContractInstance({ aci, contractAddress 
 
 ## 4. Deploy the contract
 
-If you have a Sophia contract that looks like this:
+If you have a Sophia contract source code that looks like this:
 ```sophia
 contract Increment =
 
@@ -135,7 +135,7 @@ console.log(contractInstance.deployInfo) // { owner, transaction, address, resul
 
 Note:
 
-- Deployment is only possible if the contract instance was initialized by providing sourcecode or bytecode.
+- Deployment is only possible if the contract instance was initialized by providing source code or bytecode.
 - The `init` entrypoint is a special function which is only called once for deployment, initializes the contract's state and doesn't require the `stateful` declaration.
 - In Sophia all `public functions` are called `entrypoints` and need to be declared as `stateful`
 if they should produce changes to the state of the smart contract, see `increment(value: int)`.


docs/guides/contracts.md
```diff
@@ -52,19 +52,20 @@ Note:
...
 
-### By sourcecode
+### By source code
 
 ```js
-const CONTRACT_SOURCE = ... // source code of the contract
-const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE })
+const CONTRACT_SOURCE_CODE = ... // source code of the contract
+const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE_CODE.trim() })
+// trim function here is a helper to avoid indentation problems when you directly use source code
```
janmichek, line 60: When I was redoing tutorials, I was struggling with formatting.
Then I found this hint in codebase which ultimately solved my problem.
https://github.com/aeternity/aepp-sdk-js/blob/40cc9337888454ca1d8c0e8fd25c013ae3497834/examples/browser/aepp/src/Contracts.vue#L88
So I thought we can pass this hint to other developers .]

davidyuk (commented):
> This page seems to me a bit more docs than practical example.

I agree, currently it looks like a mix of a guide and API reference. We can make it clearer by extracting some stuff to API reference (#1318), leaving links to it in a guide

docs/guides/contracts.md
janmichek: I see that using `.trim()` is not necessary.  But I got stuck for a while on this when I was redoing the tutorial.
In some tutorials was not needed to trim.


 My motivation is to make redoing examples a smooth experience.
 So adding `trim` is more about being on the safe side.
Somehow we should pass this knowledge to user.

What do you suggest then?

docs/guides/contracts.md
```diff
@@ -52,19 +52,20 @@ Note:
 
 ## 3. Initialize the contract instance
 
-### By sourcecode
+### By source code
 
 ```js
-const CONTRACT_SOURCE = ... // source code of the contract
-const contractInstance = await aeSdk.getContractInstance({ source: CONTRACT_SOURCE })
+const CONTRACT_SOURCE_CODE = ... // source code of the contract
```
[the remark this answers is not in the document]
janmichek: Thanks, agree on this. 
Renamed to  `sourceCode`

docs/guides/contracts.md
janmichek: I understand better now.
You are talking about loading from file / directly inserted contract source code to js.

But the question is still the same? Any idea how would we elegantly advise users in order to prevent them from being stuck?

janmichek:
> 

Sweet. Good to know, thanks for pointing on that issue. I didn't know this thing is coming.
Should I revert some of those changes when I know now there will be solid docs?
Yes, having separate Docs from Examples would solve much confusion around.

janmichek:
There are some tests failing. I have not changed a single line of functional code :D 
Any idea why this is happening? I will try to re-run at least

docs/guides/contracts.md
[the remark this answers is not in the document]
davidyuk: I meant to rename `source` argument of `getContractInstance` to `sourceCode`, does it makes sense? Otherwise to keep `CONTRACT_SOURCE`

docs/guides/contracts.md
davidyuk: In which case user would stuck without `trim`?
As I see the contract defined as:
```
const contractSource = `

     contract Test =
         entrypoint getArg(x : map(string, int)) = x
         entrypoint getArg2(x : map(string, int)) = x

`;
```
works fine without `trim`

davidyuk:
SDK tests are not so deterministic as we want them to be 😃

docs/guides/contracts.md
janmichek: I think it was the case in directly entered contract code in `4. Deploy the contract`. You have to manually adjust white spacing it indenting.
There is the usage of trim function in the codebase in more places for some reason, so I thought we could pass this knowledge to developers.
I have removed the trim function then

docs/guides/contracts.md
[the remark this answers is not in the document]
janmichek: The `source` does not really express what it is.
In the context of this guide `Source` can be directly pasted code, contract address, or file from the filesystem. That's what is confusing on the meaning of the word `source`
`Source code` is just without a doubt what it is. Maybe you as an experienced developer just like shorter naming for daily used tools, but  this was confusing to me when I saw it for the first time.

Should I also change the attribute name of `getContractInstance`? I didn't want to make some breaking changes.

davidyuk (approved):

janmichek:
@davidyuk Excellent, thanks for handling this!

PR #1453, opened 2022-03-31
test(examples): on prepublish

### LINKED ISSUE #445: Write smoke tests of examples
Restructuring docs, made it obvious that we need to:

- [x] correct these examples:
  1. [x] https://github.com/aeternity/aepp-sdk-js/blob/chore/new-docs/examples/node/aecontract.js
  2. [x] https://github.com/aeternity/aepp-sdk-js/blob/chore/new-docs/examples/node/aewallet.js
- [x] create new ones
- [ ] create tests for all of them (to make sure they are working when releasing)
- [x] check possibility to have these tests valid also for Browser bundles

davidyuk:
closes #445

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ coverage.*
 .nyc_output/
 /examples/**/package-lock.json
 /examples/**/dist
+/examples/**/artifacts
 .history
 .site
 site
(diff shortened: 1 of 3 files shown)


Kalovelo (approved):
Seems great, not sure if we should create separate tests for their result values

davidyuk:
It is mostly just to close the stale issue. I would continue to improve it only if we'll face any significant problems with it

Issue #1216, opened 2021-06-10
Statically generate node and compiler APIs from Swagger files


davidyuk:
Like it is done in go-sdk. This would help to
- create a TypeScript version of it (needed for #1215)
- make sdk to init in sync (#762)

davidyuk:
So, my thoughts on this:
- consider merging https://github.com/aeternity/aepp-sdk-js/pull/1368 before, because this is overlapping refactoring
- would be easier to generate API for the compiler firstly because there is no need to handle numbers (should be easy to do with approach from https://github.com/aeternity/aepp-sdk-js/pull/1233)

PR #1456, opened 2022-04-01
Generate compiler api in TypeScript using autorest

davidyuk:
fixes partially #1216 
found issues: https://github.com/aeternity/aesophia_http/issues/87 https://github.com/aeternity/aesophia_http/issues/88 https://github.com/aeternity/aesophia_http/issues/90 https://github.com/aeternity/aesophia_http/issues/91

### DIFF
--- a/src/contract/aci.js
+++ b/src/contract/aci.js
@@ -54,7 +54,7 @@ import { hash } from '../utils/crypto-ts'
  * @param {String} [options.bytecode] Contract bytecode
  * @param {Object} [options.aci] Contract ACI
  * @param {String} [options.contractAddress] Contract address
- * @param {Object} [options.filesystem] Contact source external namespaces map
+ * @param {Object} [options.fileSystem] Contact source external namespaces map
  * @param {Boolean} [options.validateBytecode] Compare source code with on-chain version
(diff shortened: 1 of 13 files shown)


src/contract/aci.js
```diff
@@ -73,12 +73,12 @@ export default async function getContractInstance ({
...
   aci: _aci,
   contractAddress,
-  filesystem = {},
+  fileSystem = {},
   validateBytecode,
   ...otherOptions
 } = {}) {
   if (!_aci && source) {
-    _aci = await this.compilerApi.generateACI({ code: source, options: { filesystem } })
+    _aci = await this.compilerApi.generateACI({ code: source, options: { fileSystem } })
```
davidyuk, line 81: for some reason sdk and aesophia_http names this option differently

Kalovelo:
Apart from the need of rebasing, I think it looks great!



I could also suggest updating the testing docs so contributors know that they need to run `build:api` before running tests. Or, we could create a new npm command which would build the api if it is not already built. We could then append this command in the test command, like



"test": "checkApi && nyc mocha './test/unit/' './test/integration/'",



Finally, since `api` folder is auto-generated through an npm command, we could consider adding it in .gitignore

PR #1487, opened 2022-04-22
Combine derivation errors to a single class, add Bip32Path type

davidyuk:
I'm sure that `NotHardenedSegmentError` and `UnsupportedChildIndexError` should be combined because they actually refers to the same issue. But do we need to be able to distinglish between invalid derivation path root and not hardened segment in the path on the level of classes? We already have a lot of error classes, I don't think that we need 3 of them for derivation.

### DIFF
--- a/src/utils/hd-wallet.ts
+++ b/src/utils/hd-wallet.ts
@@ -3,9 +3,14 @@ import { full as hmac } from 'tweetnacl-auth'
 import { fromString } from 'bip32-path'
 import { decryptKey, encryptKey } from './crypto'
 import { encode } from './encoder'
-import {
-  InvalidDerivationPathError, NotHardenedSegmentError, UnsupportedChildIndexError
-} from './errors'
+import { CryptographyError } from './errors'
+
+export class DerivationError extends CryptographyError {
+  constructor (message: string) {
+    super(message)
+    this.name = 'DerivationError'
+  }
+}
 
 const ED25519_CURVE = Buffer.from('ed25519 seed')
 const HARDENED_OFFSET = 0x80000000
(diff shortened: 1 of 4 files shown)


src/utils/hd-wallet.ts
```diff
@@ -3,9 +3,14 @@ import { full as hmac } from 'tweetnacl-auth'
...
-  InvalidDerivationPathError, NotHardenedSegmentError, UnsupportedChildIndexError
-} from './errors'
+import { CryptographyError } from './errors'
+
+export class DerivationError extends CryptographyError {
+  constructor (message: string) {
+    super(message)
+    this.name = 'DerivationError'
+  }
+}
```
davidyuk, line 13: This is a new approach, I'm proposing to store errors related to specific functionality in the same file with it. This would allow to implement complex error classes without cycle dependencies. `DerivationError` would be still exported in the root of package because of #1058

test/unit/hd-wallet.ts
```diff
@@ -158,7 +158,7 @@ describe('hd wallet', () => {
         prv: Buffer.from('551d333177df541ad876a60ea71f00447931c0a9da16f227c11ea080d7391b8d', 'hex'),
         pub: Buffer.from('0047150c75db263559a70d5778bf36abbab30fb061ad69f69ece61a72b0cfa4fc0', 'hex')
       }]
-    }]
+    }] as const)
       .forEach(({ seed, tests }, idx) => {
         describe(`vector ${idx + 1}`, () =>
           tests.forEach(test =>
```
subhod-i, line 164: a bit surprised here how the compiler is picking up the` test.path` type without `as`.

subhod-i (approved):

src/utils/hd-wallet.ts
subhod-i: Got it. Types, constants, and errors will fall into the same file if referenced only in the same file.

PR #1483, opened 2022-04-21
Support CERES protocol version

fabiankrol:
This PR adds support for the experimental protocol version used by Hyperchain Alpha.

PR #1489, opened 2022-04-25
feat: support CERES protocol version

davidyuk:
https://github.com/aeternity/aeternity/releases/tag/v0.0.0-hc5
based on #1483

### DIFF
--- a/docker/aeternity_node_mean16.yaml
+++ b/docker/aeternity_node_mean16.yaml
@@ -20,7 +20,7 @@ chain:
   persist: false
   hard_forks:
     "1": 0
-    "5": 1
+    "6": 1
 
 mining:
   autostart: true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v2
         with:
-          node-version: 16.x
+          node-version: 18.x
       - uses: actions/cache@v3
         with:
           path: |
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,7 +58,7 @@
         "@types/sinon": "^10.0.13",
         "@typescript-eslint/eslint-plugin": "^5.47.1",
         "@typescript-eslint/parser": "^5.47.1",
-        "autorest": "^3.6.2",
+        "autorest": "^3.6.3",
         "babel-loader": "^9.1.0",
         "babel-plugin-add-import-extension": "^1.6.0",
         "babel-plugin-import-globals": "^2.0.0",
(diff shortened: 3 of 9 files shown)


docker/aeternity_node_mean16.yaml
```diff
@@ -20,7 +20,7 @@ chain:
   persist: false
   hard_forks:
     "1": 0
-    "5": 1
```
subhod-i, line 23: @davidyuk How these are numbered?

davidyuk: the key is a version of consensus protocol and the value is the height at which protocol should be applied, so I'm switching our node for integration tests from 5th propcol to 6th, it is strange that nothing broke

subhod-i: Nice, Thanks.

src/node.js
```diff
@@ -71,6 +71,7 @@ const Node = AsyncInit.compose({
     })
     this.version = client.spec.info.version
     if (
+      !this.version.startsWith('0.0.0-hc') &&
```
dincho, line 74: Isn't it better the clients to bypass the check with `ignoreVersion` instead of this ugly looking hardcoded version/tag bypass ?

.env
```diff
@@ -1,2 +1,2 @@
-NODE_TAG=v6.4.0
+NODE_TAG=v0.0.0-hc5
```
dincho, line 1: is this temporary to run the tests ?

docker/aeternity_node_mean16.yaml
dincho: > it is strange that nothing broke

that means the APIs are backward compatible and most of the consensus I guess

dincho: also keep in mind that HC branch and `v0.0.0-hc5` respectively somehow behind master

loxs:
Could someone rebase this, it has conflicts currently? 
I will test it for the hyperchains UI i'm developing.

.env
velzevur: I'd stick to official node releases instead.

src/node.js
davidyuk: I would replace it with a specific version check instead of `startsWith`. I think it is fine to keep it as far as sdk being compatible with this version.

loxs:
This seems to work and I suggest that it gets merged

davidyuk:
How I can test it to ensure that it works?

loxs:
> How I can test it to ensure that it works?

Ah right... you can't at the moment, hopefully we'll have a public testnet that allows this in a few days, will update you.

loxs:
There is now a publicly available network that can be used to test this branch. The node is: https://testnet-hc.aeternity.io/
Also there is an aepp that uses this branch: https://hc-ui.dev.aepps.com/
It currently uses my fork as a dependency but I wish to switch it to this repo.
The code is here: https://github.com/aeternity/aepp-hc-ui

PR #1492, opened 2022-04-25
refactor(utils): migrate json-big.js to TS

### LINKED ISSUE #1491: migrate utils/json-big.js to TS

martinkaintas:
closes #1491 

This PR migrates the `json-big.js` util to TS.

### DIFF
--- a/src/utils/json-big.ts
+++ b/src/utils/json-big.ts
@@ -4,15 +4,15 @@ import { mapObject } from './other'
 
 const jsonBig = JsonBig({ storeAsString: true })
 
-const convertValuesToBigNumbers = value => {
+const convertValuesToBigNumbers = (value: any): any => {
   if (typeof value === 'object' && value !== null && value.constructor === Object) {
     return mapObject(value, ([key, value]) => [key, convertValuesToBigNumbers(value)])
   }
   if (Array.isArray(value)) {
     return value.map(item => convertValuesToBigNumbers(item))
   }
-  if (typeof value === 'string' && BigNumber(value).toString(10) === value) {
-    const bn = BigNumber(value)
+  if (typeof value === 'string' && new BigNumber(value).toString(10) === value) {
+    const bn = new BigNumber(value)
     bn.toJSON = () => bn.toString(10)
     return bn
   }
@@ -20,6 +20,7 @@ const convertValuesToBigNumbers = value => {
 }
 
 export default {
-  stringify: (object, ...args) => jsonBig.stringify(convertValuesToBigNumbers(object), ...args),
+  stringify: (...args: Parameters<typeof JsonBig['stringify']>): string =>
+    jsonBig.stringify(convertValuesToBigNumbers(args[0]), ...args.slice(1)),
   parse: jsonBig.parse
 }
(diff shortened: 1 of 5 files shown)


src/utils/json-big.ts
```diff
@@ -0,0 +1,62 @@
...
+    data?: {
+      channel_id?: string
+      error_code?: number
+      error_msg?: string
+      round?: number
+    }
+  }
+  id?: string
+  version?: number
+}
```
davidyuk, line 34: JSON converter shouldn't be aware of aeternity-specific types, I don't think this is needed

davidyuk, line 34: JSON converter shouldn't be aware of aeternity-specific types, I don't think this is needed

src/utils/json-big.ts
```diff
@@ -0,0 +1,62 @@
...
+    bn.toJSON = () => bn.toString(10)
+    return bn
+  }
+  return value
+}
+
+export default {
+  stringify: (
+    object: ConvertedValueType,
+    ...args: [replacer: Function | string[], space: string | number]
```
davidyuk, line 59: I propose to refer to `jsonBig.stringify` type as far as we just proxying these arguments

davidyuk, line 59: I propose to refer to `jsonBig.stringify` type as far as we just proxying these arguments

src/utils/json-big.ts
```diff
@@ -0,0 +1,62 @@
...
+      error_code?: number
+      error_msg?: string
+      round?: number
+    }
+  }
+  id?: string
+  version?: number
+}
+
+type ConvertedValueType = BigNumValueObject | BigNumber | string
```
davidyuk, line 36: This is not a complete set of types, for example, object that is converted to string can have a number in a field. In general I wouldn't limit the type of first argument of JSON.stringify (is it just `any`?).

davidyuk, line 36: This is not a complete set of types, for example, object that is converted to string can have a number in a field. In general I wouldn't limit the type of first argument of JSON.stringify (is it just `any`?).

src/utils/json-big.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const convertValuesToBigNumbers = (
+  value: ConvertedValueType): ConvertedValueType | {} => {
```
davidyuk, line 39: Why ` | {}` is needed?

src/utils/json-big.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (Array.isArray(value)) {
+    return value.map(item => convertValuesToBigNumbers(item))
+  }
+  if (typeof value === 'string' && new BigNumber(value).toString(10) === value) {
+    const bn = new BigNumber(value)
+    bn.toJSON = () => bn.toString(10)
+    return bn
+  }
+  return value
+}
```
davidyuk, line 54: I think a simple replacer can be used here instead, but anyway we will drop this code soon
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter

davidyuk (approved):

PR #1498, opened 2022-05-03
Feature/ts tx builder

### LINKED ISSUE #1488: Migrate tx/builder module to ts
Migrate `tx/builder` files to TypeScript

Kalovelo:
Closes #1488 

This PR migrates the tx builder module to ts.
Motivation behind typing for each transaction type is that those interfaces will be used through other modules such as contract


src/tx/builder/field-types.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @class
+ * @extends Field
+ */
+export class Name extends Field {
+  /**
+   * Serialize AENS name to Buffer
+   * @function
+   * @static
+   * @param {String} value - AENS name
+   * @returns {Buffer}
```
davidyuk, line 47: I'm not sure about repeating here, can `@function`, `@static`, `{String}`, `{Buffer}` be derived from TypeScript, to don't define the same two times? Is this definition for IDEs, docs generation, or to be read by developer? I assume in first two cases, TS definitions can be used instead of duplicating jsdoc properties

Kalovelo:
Closing this since its integrated into #1498

Issue #907, opened 2020-02-19
Is it necessary to call `shareWalletInfo` all the time?


davidyuk:
Right now, the example wallet sends to aepp `walletInfo` messages 5 times with an interval of 5 seconds. Are these messages necessary for aepp to work? If yes then in general case, it seems not enough because we can't track navigations in an iframe, so the user might switch between aepps at any time, which makes the wallet to poll iframe with `walletInfo` messages all the time. Can aepp initialize the connection to the wallet by itself as it was before?

davidyuk:
Also, can this polling part be moved to the sdk side?

davidyuk:
Should interval be decreased to something like 1 second to speedup wallet detection?

nduchak:
> Should interval be decreased to something like 1 second to speedup wallet detection?

you have full control over this

nduchak:
> Are these messages necessary for aepp to work? If yes then in general case, it seems not enough because we can't track navigations in an iframe, so the user might switch between aepps at any time, which makes the wallet to poll iframe with walletInfo messages all the time

We can track the navigation in Aepp stamp and then trigger the onDisconnect message to wallet where you can start control polling.
The things is that one approach should work the same for all of the cases(extension, iframe/reverse iframe), and for extension wallet it's not possible to AEPP connection initialization because the `runtime.id` is required for that

davidyuk:
> you have full control over this

Have you tried to find the best decision? It looks like a controversial issue caused by a wrong architectural decision on the sdk side.

> We can track the navigation in Aepp stamp and then trigger the onDisconnect message to wallet where you can start control polling.

What if user switcher from a usual web site to aepp? This can happen at an arbitrary moment, can we track it somehow without polling?

> The things is that one approach should work the same for all of the cases

why we can't use a bit different (and much more efficient) protocol in the iframe case? Anyway, aepp sees the difference between sending messages to a parent frame or to an extension.

davidyuk:
I want to notice that the wallet team considered default behavior as a bug and has switched to infinite pooling. I'm proposing to update the sdk examples accordingly.

gorbak25:
@davidyuk Is there a chance to make any progress on this issue in the near future? There are a lot of aepps where it's clear that something's wrong:
- Logging to jelly swap takes long using SH
- People complain on the forum: https://forum.aeternity.com/t/most-dapps-running-on-aeternity-can-t-log-in-smoothly/8130/5
- etc...

[...]

PR #1502, opened 2022-05-05
Web wallet extension: shouldn't open connections to pages that not using it 

### LINKED ISSUE #1417: Browser extension shouldn't open connections to pages that not using it
https://github.com/aeternity/aepp-sdk-js/blob/9f2f22a97c84b00efda5f11373836eddfaa26f50/examples/browser/wallet-web-extension/src/content-script.js#L19

https://github.com/aeternity/aepp-sdk-js/blob/9f2f22a97c84b00efda5f11373836eddfaa26f50/examples/browser/wallet-web-extension/src/background.js#L67-L73

Together these parts make all opened pages be in `sdk.rpcClients`. Superhero wallet reloads connected pages in some cases and this makes it reload extra pages.

Try to make the fix not breaking aepp<->content-script connection.

subhod-i:
Closes #1417, #907 

Solution: is to move ` const port = browser.runtime.connect()` this from content script to content script bridge.

Side effects: Aepps will work without any change to the existing codebase.

Implementation: The `scan` method from the wallet detector sends an additional request via postMessage to the window. The content script bridge will then will establish the bridge between the page and the wallet. Now the wallet will not simply connect to all the tabs open in the browser instead opens connections to the aepps only when requested(via wallet detector). This fix also stops the continuous polling of wallet information to the tabs. A lot of CPU cycles are being saved since the wallet connects to only requested pages.

Fixed encoder type errors too.

### DIFF
--- a/examples/browser/wallet-web-extension/src/background.js
+++ b/examples/browser/wallet-web-extension/src/background.js
@@ -73,7 +73,6 @@ import {
     aeSdk.addRpcClient(connection)
     // share wallet details
     aeSdk.shareWalletInfo(port.postMessage.bind(port))
-    setInterval(() => aeSdk.shareWalletInfo(port.postMessage.bind(port)), 3000)
   })
 
   console.log('Wallet initialized!')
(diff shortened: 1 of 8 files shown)


subhod-i:
Closing in favor of https://github.com/aeternity/aepp-sdk-js/pull/1578

PR #1505, opened 2022-05-05
refactor: migrate node-pool module to TS

### LINKED ISSUE #1504: Migrate node-pool module to ts

Kalovelo:
closes #1504 
Migrates the node-pool mode to ts. Some ts-expect-error were necessary since this module is dependent on the Node module.


src/node-pool/index.ts
```diff
@@ -0,0 +1,164 @@
...
+type Node = any
+
+interface NodeInfo {
+  name: string
+  version: string
+  networkId: string
+  consensusProtocolVersion: string
+}
+
+interface NodePool {
```
davidyuk, line 22: I didn't get why do we need a separate interface, can we use something like `InstanceType<typeof  _NodePool>` instead?

src/node-pool/index.ts
```diff
@@ -0,0 +1,164 @@
...
+            throw new NodeNotFoundError('You can\'t use Node API. Node is not connected or not defined!')
+          }
+          return this.selectedNode.instance.api
+        }
+      }
+    })
+  }
+
+  /**
+     * Add Node
```
davidyuk, line 63:
```suggestion
   * Add Node
```

src/node-pool/index.ts
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Node Pool Stamp
+ * This stamp allow you to make basic manipulation(add, remove, select) on list of nodes
+ * @alias module:@aeternity/aepp-sdk/es/node-pool
+ * @param options - Initializer object
+ * @param options.nodes - Array with Node instances
+ * @return NodePool instance
+ */
+export default stampit<NodePool>({
+  init: _NodePool.prototype.init as any,
```
davidyuk, line 155: is `as any` required?

src/node-pool/index.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+     * Get array of available nodes
+     * @alias module:@aeternity/aepp-sdk/es/node-pool
+     * @example
+     * nodePool.getNodesInPool()
+     */
+  getNodesInPool (): NodeInfo[] {
+    return Array.from(this.pool.entries()).map(([name, node]: [string, Node]) => ({
```
davidyuk, line 139:
```suggestion
    return Array.from(this.pool.entries()).map(([name, node]) => ({
```
we can omit type in this case, because it can be derived from the pool type (also linter don't forces us to do that)

davidyuk (approved):

PR #1510, opened 2022-05-11
refactor(tx): migrate Tx module to ts

### LINKED ISSUE #1509: Migrate Tx module to TS

Kalovelo:
closes #1509 , #1498 

This PR migrates tx module to TypeScript.


src/tx/builder/constants.ts
```diff
@@ -88,7 +92,9 @@ export const ID_TAG = {
   channel: ID_TAG_CHANNEL
 }
 
-export const PREFIX_ID_TAG = {
+export const PREFIX_ID_TAG: {
+  [prefix: string]: number
```
davidyuk, line 96: This actually make the type less precise, I proposing to add `as const` at the end instead. Then the strings defined inside would be checkable by type system.

src/tx/builder/field-types.ts
```diff
@@ -0,0 +1,134 @@
...
+    return Buffer.from(value)
+  }
+
+  /**
+   * Deserialize value from Buffer
+   * @function
+   * @param {Buffer} value
+   * @return {String}
+   */
+  static deserialize (value: Buffer): string {
```
davidyuk, line 31:
```suggestion
  static deserialize (value: Buffer): any {
```
this is a general class and its inheritors can return actually anything, the same with arguments of `serialize`

src/tx/builder/field-types.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Serialize AENS name to Buffer
+   * @param value - AENS name
+   */
+  static serialize (value: string): Buffer {
+    ensureNameValid(value)
+    return Buffer.from(value)
+  }
+
+  /**
+   * Deserialize AENS name from Buffer
```
davidyuk, line 52: These comments is not very useful, I don't think we will have something special here to explain, otherwise we can generate they by a pattern 😛

src/tx/builder/field-types.ts
```diff
@@ -0,0 +1,134 @@
...
+ * @class
+ * @extends Field
+ */
+export class Name extends Field {
+  /**
+   * Serialize AENS name to Buffer
+   * @param value - AENS name
+   */
+  static serialize (value: string): Buffer {
+    ensureNameValid(value)
```
davidyuk, line 47:
```suggestion
  static serialize (value: `${string}.chain`): Buffer {
```
but `${string}.chain` should be extracted somewhere

src/tx/builder/index.ts
```diff
@@ -0,0 +1,743 @@
...
+ * JavaScript-based Transaction builder
+ * @module @aeternity/aepp-sdk/es/tx/builder
+ * @example import { calculateFee } from '@aeternity/aepp-sdk'
+ */
+
+/**
+ * @interface TxOptionsCommon
+ * These options are common and can be provided to every tx-type.
+ */
+interface TxOptionsCommon {
```
davidyuk, line 48: I'm not sure that we will have a general options at the end, I would combine options of child method calls instead

src/tx/builder/index.ts
```diff
@@ -0,0 +1,743 @@
...
+   * The TTL of the oracle that defines its expiration.
+   * @default {type:'delta', value:500}
+   */
+  oracleTtl?: {
+    type?: 'delta' | 'block'
+    value?: number
+  }
+}
+
+interface TxOptionsOracleQuery extends TxOptionsCommon {
```
davidyuk, line 194: The interfaces like this one I assume should be generated by schema (see my example in the last commit)

src/tx/builder/index.ts
```diff
@@ -0,0 +1,743 @@
...
+ * @param [options={}] options
+ * @param [options.excludeKeys] Array of keys to exclude for validation and build
+ * @param [options.prefix] Prefix of transaction
+ * @throws {Error} Validation error
+ * @returns object
+ * @returns object.tx Base64Check transaction hash with 'tx_' prefix
+ * @returns object.rlpEncoded rlp encoded transaction
+ * @returns object.binary binary transaction
+ */
+export function buildTx<Tx extends TxParams> (
```
davidyuk, line 665: I propose to generate parameters based on schema, and to choose the right parameters set depending on `type` value

src/tx/builder/index.ts
```diff
@@ -0,0 +1,743 @@
...
+    { excludeKeys, denomination: ('denomination' in params ? params.denomination : undefined) ?? denomination }
+  ).filter((e?: Buffer) => e != null)
+
+  const rlpEncoded = Buffer.from(rlpEncode(binary))
+  const tx: EncodedData<'tx'> = encode<'tx'>(rlpEncoded, prefix as 'tx')
+
+  return { tx, rlpEncoded, binary, txObject: unpackRawTx<Tx>(binary, schema) }
+}
+
+export interface TxHashUnpacked<Tx extends TxParams> {
```
davidyuk, line 683:
```suggestion
export interface TxUnpacked<Tx extends TxParams> {
```
hash can't be unpacked, it is one-way thing

src/tx/builder/schema.ts
```diff
@@ -403,15 +509,47 @@ const NAME_TRANSFER_TX = [
...
+  // ...BASE_TX,
+  ['tag', FIELD_TYPES.int],
+  ['VSN', FIELD_TYPES.int],
+  ['accountId', FIELD_TYPES.id, 'ak'],
+  ['nonce', FIELD_TYPES.int],
+  ['nameId', NameId],
+  ['fee', FIELD_TYPES.int],
+  ['ttl', FIELD_TYPES.int]
+] as const
+export interface TxNameRevoke extends BuildTxArgBySchema<typeof NAME_REVOKE_TX> {}
```
davidyuk, line 542: This example is for buildTx, the similar thing needs to be done for unpackTx, it can't be the same because, for example, as `amount` we accepting `string | number | BugNumber`, but returning the only `string`.
Also the code is broken now, because other parts should be rewrited as well.

src/tx/builder/schema.ts
```diff
@@ -321,10 +363,19 @@ export const TX_FEE_OTHER_GAS = (txType, txSize, { relativeTtl, innerTxSize }) =
   }
 }
 
+export interface TxBase {
+  tag?: number
+  vsn?: number
+  gasLimit?: number
+  fee?: BigNumber | number | string
+  ttl?: BigNumber | number | string
+  nonce?: number | BigNumber | string
```
davidyuk, line 372: Not sure why the base interface have more fields than the base schema (defined below)

src/tx/validator.ts
```diff
@@ -33,27 +65,41 @@ const validators = [
...
     }]
   },
-  ({ encodedTx, tx }, { node, parentTxTypes, txType }) => {
+  ({
+    encodedTx,
+    tx
+  }: {
+    encodedTx: EncodedTx<TxParams>
+    tx: TxSigned
+  }, {
```
davidyuk, line 74: Here should be a common type to extract, all validators called with the same arguments

davidyuk (changes requested):
I think we need to merge #1519 before and to rebase this one at the top of it. I've removed a lot of code there and it should make easier to finish this PR then

src/tx/builder/helpers.ts
```diff
@@ -32,58 +32,63 @@ export const createSalt = salt
...
- * @param {string} ownerId The public key of the owner account
- * @param {number} nonce the nonce of the transaction
- * @return {string} Contract public key
+ * @param ownerId The public key of the owner account
+ * @param nonce the nonce of the transaction
+ * @return Contract public key
  */
-export function buildContractId (ownerId, nonce) {
-  const ownerIdAndNonce = Buffer.from([...decode(ownerId, 'ak'), ...toBytes(nonce)])
+export function buildContractId (ownerId: string, nonce: number | BigNumber): string {
```
davidyuk, line 47:
```suggestion
export function buildContractId (ownerId: EncodedData<'ak'>, nonce: number | BigNumber): EncodedData<'ct'> {
```

src/tx/builder/helpers.ts
```diff
@@ -32,58 +32,63 @@ export const createSalt = salt
...
+ * @alias module:@aeternity/aepp-sdk/es/tx/builder/helpers
+ * @param senderId The public key of the sender account
+ * @param nonce the nonce of the transaction
+ * @param oracleId The oracle public key
+ * @return Contract public key
  */
-export function oracleQueryId (senderId, nonce, oracleId) {
-  function _int32 (val) {
+export function oracleQueryId (
+  senderId: string,
```
davidyuk, line 62:
```suggestion
  senderId: EncodedData<'ak'>,
```
?

src/tx/builder/helpers.ts
```diff
@@ -32,58 +32,63 @@ export const createSalt = salt
...
+  nonce: number | BigNumber | string,
+  oracleId: EncodedData<'ok'>): string {
```
davidyuk, line 64:
```suggestion
  oracleId: EncodedData<'ok'>
): string {
```

src/tx/builder/helpers.ts
```diff
@@ -32,58 +32,63 @@ export const createSalt = salt
...
+  function _int32 (val: number): Buffer {
     const nonceBE = toBytes(val, true)
     return concatBuffers([Buffer.alloc(32 - nonceBE.length), nonceBE])
   }
 
-  const b2bHash = hash(Buffer.from([...decode(senderId, 'ak'), ..._int32(nonce), ...decode(oracleId, 'ok')]))
+  const b2bHash = hash(Buffer.from([...decode(senderId as EncodedData<'ak'>), ..._int32(nonce as number), ...decode(oracleId)]))
```
davidyuk, line 70:
```suggestion
  const b2bHash = hash(Buffer.from([...decode(senderId as EncodedData<'ak'>), ..._int32(nonce), ...decode(oracleId)]))
```
should we update `toBytes` definition instead?

src/tx/builder/helpers.ts
```diff
@@ -32,58 +32,63 @@ export const createSalt = salt
...
- * @function
+ *
  * @alias module:@aeternity/aepp-sdk/es/tx/builder/helpers
- * @param {String} name Name to encode
- * @return {String} `nm_` prefixed encoded AENS name
+ * @param name Name to encode
+ * @return `nm_` prefixed encoded AENS name
  */
-export function produceNameId (name) {
+export function produceNameId (name: string): EncodedData<'nm'> {
```
davidyuk, line 91:
```suggestion
export function produceNameId (name: `${string}.chain`): EncodedData<'nm'> {
```

src/tx/builder/helpers.ts
```diff
@@ -194,25 +189,20 @@ const AENS_SUFFIX = '.chain'
...
  * @alias module:@aeternity/aepp-sdk/es/tx/builder/helpers
- * @param {string} name
- * @return void
+ * @param name
  * @throws Error
  */
-export function ensureNameValid (name) {
-  if (!name || typeof name !== 'string') throw new InvalidNameError('Name must be a string')
+export function ensureNameValid (name: string): void {
   if (!name.endsWith(AENS_SUFFIX)) throw new InvalidNameError(`Name should end with ${AENS_SUFFIX}: ${name}`)
```
davidyuk, line 197: This check can be replaced with TypeScript

src/tx/builder/helpers.ts
```diff
@@ -223,55 +213,62 @@ export function isNameValid (name) {
...
- * @returns {String} default AENS pointer key
+ * @returns default AENS pointer key
  * @throws exception when default key not defined
  */
-export function getDefaultPointerKey (identifier) {
-  decode(identifier)
-  const prefix = identifier.substr(0, 2)
-  return POINTER_KEY_BY_PREFIX[prefix] ||
-    (() => { throw new NoDefaultAensPointerError(prefix) })()
+export function getDefaultPointerKey (identifier: EncodedData<any>): string {
```
davidyuk, line 219:
```suggestion
export function getDefaultPointerKey (identifier: EncodedData<keyof POINTER_KEY_BY_PREFIX>): POINTER_KEY_BY_PREFIX[keyof POINTER_KEY_BY_PREFIX] {
```
🤔

src/tx/builder/index.ts
```diff
@@ -0,0 +1,569 @@
...
+/**
+ * Calculate fee based on tx type and params
+ * @param txType
+ * @param options.params
+ * @param options.multiplier
+ * @param options.vsn
+ * @return {BigNumber}
+ */
+function buildFee (txType: TxType, { params, multiplier, vsn }:
+{params: any, multiplier: BigNumber, vsn?: number}): BigNumber {
```
davidyuk, line 317:
```suggestion
function buildFee (
  txType: TxType,
  { params, multiplier, vsn }: { params: any, multiplier: BigNumber, vsn?: number }
): BigNumber {
```
messing at the same line function name, parameters and return type looks difficult to read

test/integration/channel.ts
```diff
@@ -768,7 +766,7 @@ describe('Channel', function () {
       poi,
       payload: signedTx
     })
-    const closeSoloTxFee = unpackTx(closeSoloTx).tx.fee
+    const closeSoloTxFee = unpackTx<TxChannelCloseSolo>(closeSoloTx).tx.fee
```
davidyuk, line 769:
```suggestion
    const closeSoloTxFee = unpackTx(closeSoloTx, TX_TYPE.channelCloseSolo).tx.fee
```
So, what about doing a runtime check? And switching the return type accordingly?

src/tx/builder/schema.ts
```diff
@@ -317,477 +359,578 @@ export const TX_FEE_OTHER_GAS = (txType, txSize, { relativeTtl, innerTxSize }) =
...
 
 const ACCOUNTS_TREE_TX = [
   ...BASE_TX,
-  TX_FIELD('accounts', FIELD_TYPES.rlpBinary)
-]
+  ['accounts', FIELD_TYPES.rlpBinary]
+] as const
+export interface TxAccountsTree extends BuildTxArgBySchema<typeof ACCOUNTS_TREE_TX> {}
+
+export type TxParams = TxAccount | TxSpend | TxSigned
```
davidyuk, line 919: I don't think you need to combine these interfaces, better to switch to specific one in particular case

src/tx/builder/schema.ts
```diff
@@ -245,6 +240,29 @@ export const OBJECT_ID_TX_TYPE = {
...
+  hex: string
+  rlpBinary: any
+  rlpBinaries: any[]
+  rawBinary: string
+  signatures: Uint8Array[]
+  pointers: Pointer[]
+  offChainUpdates: any
+  callStack: any
+  proofOfInclusion: any
+  mptrees: any
```
davidyuk, line 262:
```suggestion
  mptrees: MPTree[]
```
🤔

src/tx/builder/schema.ts
```diff
@@ -245,6 +240,29 @@ export const OBJECT_ID_TX_TYPE = {
   [OBJECT_TAG_GA_META]: TX_TYPE.gaMeta,
   [OBJECT_TAG_PAYING_FOR]: TX_TYPE.payingFor,
   [OBJECT_TAG_SOPHIA_BYTE_CODE]: TX_TYPE.sophiaByteCode
+} as const
+
+interface BuildFieldTypes {
```
davidyuk, line 245: also you need a similar structure for unpacking

src/tx/builder/schema.ts
```diff
@@ -317,477 +359,578 @@ export const TX_FEE_OTHER_GAS = (txType, txSize, { relativeTtl, innerTxSize }) =
...
-  TX_FIELD('payable', FIELD_TYPES.bool)
-]
+  ['sourceCodeHash', FIELD_TYPES.rawBinary],
+  ['typeInfo', FIELD_TYPES.sophiaCodeTypeInfo],
+  ['byteCode', FIELD_TYPES.rawBinary],
+  ['compilerVersion', FIELD_TYPES.string],
+  ['payable', FIELD_TYPES.bool]
+] as const
+export interface TxContractByteCodeLima
+  extends BuildTxArgBySchema<typeof CONTRACT_BYTE_CODE_LIMA> {}
```
davidyuk, line 404: Is it actually used externally? Should be a way to don't duplicate these exports as well

davidyuk (commented):
May be resonable to extract a subset of these changes in a separate PR

test/integration/channel.ts
Kalovelo: Currently working on this, since unpackTx already accepts an optional argument `fromRlpBinary`, should I create an object argument regarding options, something like `options: {
txType?: TxType,
fromRlpBinary?: boolean
}?`

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+ * @param options.vsn
+ * @return {BigNumber}
+ */
+function buildFee (
+  txType: TxType,
+  { params, multiplier, vsn }: { params: any, multiplier: BigNumber, vsn?: number }
+): BigNumber {
+  const { rlpEncoded: txWithOutFee } = buildTx({ ...params }, txType, { vsn })
+  const txSize = txWithOutFee.length
+  const txTypes: [TxType, TxType] = [TX_TYPE.gaMeta, TX_TYPE.payingFor]
```
davidyuk, line 322:
```suggestion
  const txTypes = [TX_TYPE.gaMeta, TX_TYPE.payingFor] as const
```

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+ * @param fee - fee
+ * @param txType - Transaction type
+ * @param options - Options object
+ * @param options.gasLimit
+ * @param options.params - Tx params
+ * @example calculateFee(null, 'spendTx', { gasLimit, params })
+ */
+export function calculateFee (
+  fee: number | BigNumber | string = 0,
+  txType: TxType, { gasLimit = 0, params, showWarning = true, vsn }: {
```
davidyuk, line 346:
```suggestion
  txType: TxType,
  { gasLimit = 0, params, showWarning = true, vsn }: {
```

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+  } = {}
+): Prefix extends EncodingType
+    ? BuiltTx<TxSchema, Prefix>
+    : BuiltTx<TxSchema, 'tx'> {
+  const schemas = TX_SERIALIZATION_SCHEMA[type]
+  if (schemas == null) throw new UnexpectedTsError()
+
+  vsn ??= Math.max(
+    ...Object.keys(schemas).map(a => +a))
+  const schema = schemas[vsn as keyof typeof TX_SERIALIZATION_SCHEMA[typeof type]]
```
davidyuk, line 479: `as` part looks overcomplicated, isn't `van` just a number?

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+ * @param encodedTx String or RLP encoded transaction array
+ * (if fromRlpBinary flag is true)
+ * @param fromRlpBinary Unpack from RLP encoded transaction (default: false)
+ * @param prefix - Prefix of data
+ * @returns object
+ * @returns object.tx Object with transaction param's
+ * @returns object.rlpEncoded rlp encoded transaction
+ * @returns object.binary binary transaction
+ */
+export function unpackTx<TxType> (
```
davidyuk, line 522:
```suggestion
export function unpackTx<TxType extends keyof TxTypeSchemas> (
```
and to remove the switch at 528, not sure why it is needed

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+  { txType?: TxType, fromRlpBinary?: boolean } = {
+    fromRlpBinary: false
+  }
+): TxType extends keyof TxTypeSchemas
+    ? TxUnpacked<TxTypeSchemas[TxType]>
+    : TxUnpacked<TxSigned> {
+  const rlpEncoded = fromRlpBinary ? encodedTx as Uint8Array : decode(encodedTx as EncodedData<'tx'>)
+  const binary = rlpDecode(rlpEncoded)
+  const objId = readInt(binary[0] as Buffer) as unknown as keyof typeof TX_DESERIALIZATION_SCHEMA
+  txType ??= OBJECT_ID_TX_TYPE[objId] as unknown as TxType
```
davidyuk, line 534: Shouldn't be necessary to do `as unknown as TxType`

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+    binary,
+    txObject: unpackRawTx(binary, schema)
+  } as any
+}
+
+export interface TxUnpacked<Tx extends TxSchema> {
+  txType: TxType
+  tx: RawTxObject<Tx>
+  rlpEncoded: Uint8Array
+  binary: Uint8Array | NestedUint8Array
```
davidyuk, line 507:
```suggestion
  binary: NestedUint8Array
```
Isn't `NestedUint8Array` a superset of `Uint8Array`?

src/tx/builder/schema.ts
```diff
@@ -74,23 +77,18 @@ const OBJECT_TAG_GA_META = 81
 const OBJECT_TAG_PAYING_FOR = 82
 const OBJECT_TAG_SOPHIA_BYTE_CODE = 70
 
+export type TxField = [name: string,
+  type: string | typeof Field,
+  prefix?: EncodingType | EncodingType[]]
```
davidyuk, line 82:
```suggestion
export type TxField = [
  name: string,
  type: string | typeof Field,
  prefix?: EncodingType | EncodingType[]
]
```

src/tx/builder/schema.ts
```diff
@@ -74,23 +77,18 @@ const OBJECT_TAG_GA_META = 81
...
+
+export type OnAccount = string | {
+  publicKey: EncodedData<'ak'>
+  secretKey: string
+}
```
davidyuk, line 87: `Keypair` is already defined in `memory.ts`

src/tx/builder/schema.ts
```diff
@@ -292,11 +327,27 @@ export const TX_FEE_BASE_GAS = (txType) => {
     [TX_TYPE.gaAttach]: 5,
     [TX_TYPE.gaMeta]: 5,
     [TX_TYPE.payingFor]: 1 / 5
-  })[txType] || 1
+  }
```
davidyuk, line 330:
```suggestion
  } as const
```
and remove `[key in TxType]?: number` from the above

src/tx/builder/schema.ts
```diff
@@ -926,8 +1059,71 @@ export const TX_SERIALIZATION_SCHEMA = {
   [TX_TYPE.payingFor]: {
     1: PAYING_FOR_TX
   }
+} as const
+
+export type TxType = keyof typeof TX_SERIALIZATION_SCHEMA
+
+export interface TxTypeSchemas {
```
davidyuk, line 1066: would require less code to generate it based on `TX_SERIALIZATION_SCHEMA` or so
including dropping the similar definitions like:
```
interface TxAccountsTree extends BuildTxArgBySchema<typeof ACCOUNTS_TREE_TX> {}
```

davidyuk (changes requested):
Too many issues, I'm giving up

src/tx/builder/field-types.ts
```diff
@@ -0,0 +1,114 @@
+import {
+  writeId, readId, produceNameId, getMinimumNameFee, readInt, writeInt, isNameValid
+} from './helpers'
+import { InsufficientNameFeeError, IllegalArgumentError } from '../../utils/errors'
+import BigNumber from 'bignumber.js'
+import { MIN_GAS_PRICE } from './constants'
+
+export class Field {
+  public prototype: any
```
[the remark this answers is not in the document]
Kalovelo: Without it this error pops;

class Field
Unexpected class with only static properties.eslint[@typescript-eslint/no-extraneous-class](https://typescript-eslint.io/rules/no-extraneous-class)

src/tx/builder/index.ts
Kalovelo: It is, but schemas is an object of readonly keys. thereforce having simply as `schemas[vsn]` pops a tserror

Changing it to `vsn as keyof typeof schemas`
`

src/tx/builder/index.ts
```diff
@@ -0,0 +1,579 @@
...
+  ).filter(e => e !== undefined)
+
+  const rlpEncoded = rlpEncode(binary)
+  const tx = encode(rlpEncoded, prefix)
+  return {
+    tx,
+    rlpEncoded,
+    binary,
+    txObject: unpackRawTx(binary, schema)
+  } as any
```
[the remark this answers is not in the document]
Kalovelo: following the approach here: https://stackoverflow.com/questions/52817922/typescript-return-type-depending-on-parameter/52818072#52818072

src/tx/builder/index.ts
Kalovelo: rlp module also uses it as a Union

image

tsconfig.json
```diff
@@ -20,12 +20,7 @@
...
-    "typeRoots": [
-      "./node_modules/@types",
-      "./src/typings"
-    ]
+    "typeRoots": ["./node_modules/@types", "./src/typings"]
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*"]
```
davidyuk, line 25: please revert changes in this file

davidyuk (changes requested):

src/chain.ts
```diff
@@ -54,15 +52,15 @@ export function _getPollInterval (
...
   api: InstanceType<typeof NodeApi>
+  nodeNetworkId: string
 }
-
 interface Account {
   address: (options: any) => Promise<EncodedData<'ak'>>
 }
 
-type AensName = `${string}.chain`
+export type AensName = `${string}.chain`
```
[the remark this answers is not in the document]
Kalovelo: in this PR ae/aens is still in JS, therefore this could maybe be done in that PR.
(Impossible to have in .js files: https://stackoverflow.com/questions/53710368/how-do-i-add-typescript-types-to-a-javascript-module-without-switching-to-typesc)

src/tx/builder/index.ts
```diff
@@ -0,0 +1,583 @@
...
+  | typeof TX_TYPE.oracleResponse
+
+function getOracleRelativeTtl (params: any, txType: TxType): number {
+  const ttlKeys = {
+    [TX_TYPE.oracleRegister]: 'oracleTtl',
+    [TX_TYPE.oracleExtend]: 'oracleTtl',
+    [TX_TYPE.oracleQuery]: 'queryTtl',
+    [TX_TYPE.oracleResponse]: 'responseTtl'
+  } as const
+  const ttlKey = ttlKeys[txType as oracleTxTtlType]
```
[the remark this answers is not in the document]
Kalovelo: done in this case method, there's an issue with imported objects

test/integration/txVerification.ts
```diff
@@ -1,14 +1,17 @@
 import { before, describe, it } from 'mocha'
 import { expect } from 'chai'
+// @ts-expect-error
 import { getSdk } from '.'
 import { generateKeyPair } from '../../src/utils/crypto'
 import MemoryAccount from '../../src/account/memory'
 import verifyTransaction from '../../src/tx/validator'
-import { ArgumentError } from '../../src/utils/errors'
+import { InvalidTxParamsError } from '../../src/utils/errors'
+// @ts-expect-error
```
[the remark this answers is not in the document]
davidyuk: I meant the `TX_TYPE` from below, if it would be imported from ts file then its type will be validated

davidyuk (approved):

PR #1511, opened 2022-05-12
refactor(contract): migrate contract module to ts

### LINKED ISSUE #1507: Migrate Contract module to TS

martinkaintas:
closes #1507


src/ae/contract.ts
```diff
@@ -0,0 +1,212 @@
...
+   * Utility method to create a delegate signature for a contract
+   * @alias module:@aeternity/aepp-sdk/es/ae/contract
+   * @category async
+   * @param ids The list of id's to prepend
+   * @param opt Options
+   * @return Signature in hex representation
+   */
+  async delegateSignatureCommon (
+    ids: string[] = [],
+    opt: { onAccount?: string | object } = {}): Promise<string> {
```
davidyuk, line 51: `onAccount` is accepted by one of the function we are passing options to, don't define options by yourself, combine options types of `getNetworkId` and `sign` instead

src/ae/contract.ts
```diff
@@ -0,0 +1,212 @@
...
+   * const params = { contractId }
+   * const preclaimSig = await aeSdk.createAensDelegationSignature(params, { onAccount: current })
+   * // Claim, transfer and revoke signature
+   * const params = { contractId, name }
+   * const aensDelegationSig = await contract.createAensDelegationSignature(
+   *   params, name, { onAccount: current }
+   * )
+   */
+  async createAensDelegationSignature (
+    { contractId, name }: {contractId: string, name?: string[]},
```
davidyuk, line 87:
```suggestion
    { contractId, name }: {contractId: string, name?: string},
```
there is a single name

src/ae/contract.ts
```diff
@@ -0,0 +1,212 @@
...
+ * @returns Contract instance
+ * @example
+ * import { Transaction, MemoryAccount, ChainNode } from '@aeternity/aepp-sdk
+ *
+ * const ContractWithAE = await Contract
+ *    .compose(Transaction, MemoryAccount, ChainNode) // AE implementation
+ * const client = await ContractWithAe({ url, compilerUrl, keypair, ... }) *
+ */
+
+export default Ae.compose<_Contract>(ContractCompilerHttp, {
```
davidyuk, line 194: This module looks more like a set of function than an entity (like account), proposing to rewrite it as I did with chain module in #1233

src/ae/contract.ts
```diff
@@ -0,0 +1,212 @@
...
+    // AENS
+    createAensDelegationSignature: _Contract.prototype.createAensDelegationSignature,
+    // Oracle
+    createOracleDelegationSignature: _Contract.prototype.createOracleDelegationSignature
+  },
+  deepProps: {
+    Ae: {
+      defaults: {
+        gasPrice: MIN_GAS_PRICE,
+        amount: AMOUNT
```
davidyuk, line 208: try to inline these defaults into places where they actually needed, `AMOUNT` probably can be removed at all because it is 0

src/contract/aci.ts
```diff
@@ -0,0 +1,487 @@
...
+ */
+
+// @ts-expect-error TODO remove
+import { Encoder as Calldata } from '@aeternity/aepp-calldata'
+// @ts-expect-error TODO remove
+import { DRY_RUN_ACCOUNT, GAS_MAX } from '../tx/builder/schema'
+// @ts-expect-error TODO remove
+import TxObject from '../tx/tx-object'
+// @ts-expect-error TODO remove
+import { decode } from '../tx/builder/helpers'
```
davidyuk, line 32:
```suggestion
import { decode } from '../utils/encoder'
```

src/contract/aci.ts
```diff
@@ -0,0 +1,487 @@
...
+    name: string
+    event: any
+    kind: string
+    state: any
+    type_defs: any[]
+    functions: FunctionACI[]
+  }}
+  externalEncodedAci: any[]
+  interface: string
+}
```
davidyuk, line 70: I propose to enchance the ACI definition that compiler has in swagger file by adding a directive in https://github.com/aeternity/aepp-sdk-js/blob/777e990f2e900b4d8df8b4c4d91c1d4e9bab6f72/tooling/autorest/compiler.yaml#L17-L16, instead of doing an extra aci defnintion.
Later we will make a PR with it to compiler repo.

src/contract/aci.ts
```diff
@@ -0,0 +1,487 @@
...
+  * @return bytecode
+  */
+  instance.compile = async (options = {}): Promise<string> => {
+    if (instance.bytecode != null) throw new IllegalArgumentError('Contract already compiled')
+    if (instance.source == null) throw new IllegalArgumentError('Can\'t compile without source code')
+    const { bytecode } = await this.compilerApi.compileContract({
+      code: instance.source, options: { ...instance.options, ...options }
+    })
+    instance.bytecode = bytecode
+    return instance.bytecode ?? ''
```
davidyuk, line 235: Doesn't seems there is a case when bytecode would be empty

src/contract/aci.ts
```diff
@@ -0,0 +1,487 @@
...
+    const txData = await this.send(tx, options)
+    const result = {
+      hash: txData.hash, tx: TxObject({ tx: txData.rawTx }), txData, rawTx: txData.rawTx
+    }
+    if (txData.blockHeight == null) return result
+    const { callInfo } = await this.api.getTransactionInfoByHash(txData.hash)
+    Object.assign(result.txData, callInfo) // TODO: don't duplicate data in result
+    handleCallError(callInfo, tx)
+    const res = { ...result, result: callInfo }
+    return res
```
davidyuk, line 254:
```suggestion
    return { ...result, result: callInfo }
```

src/contract/aci.ts
```diff
@@ -0,0 +1,487 @@
...
+  * Decode Events
+  * @alias module:@aeternity/aepp-sdk/es/contract/aci
+  * @param events Array of encoded events (callRes.result.log)
+  * @param options Options
+  * @returns DecodedEvents
+  */
+  instance.decodeEvents = (
+    events: Event[],
+    { omitUnknown, ...opt }: {omitUnknown?: boolean} = {}
+  ): Array<DecodedEvent | null> => events
```
davidyuk, line 433: Actually, there can't be null in result because of the last line

test/integration/contract-aci.js
```diff
@@ -300,7 +300,7 @@ describe('Contract instance', function () {
   it('fails on paying to not payable function', async () => {
     const amount = 100
     await expect(testContract.methods.intFn.send(1, { amount }))
-      .to.be.rejectedWith(NotPayableFunctionError, `You try to pay "${amount}" to function "intFn" which is not payable. Only payable function can accept coins`)
+      .to.be.rejectedWith(NotPayableFunctionError, `You try to pay "${amount}" to function "intFn" which is not payable. Only payable function can accept tokens`)
```
davidyuk, line 303: Actually, AE should be named a coin, see #1514

src/contract/ga/index.ts
```diff
@@ -0,0 +1,266 @@
...
+// @ts-expect-error TODO: remove me
+import { buildTx, unpackTx } from '../../tx/builder'
+import { prepareGaParams } from './helpers'
+import { hash } from '../../utils/crypto'
+import { decode, EncodedData } from '../../utils/encoder'
+import { IllegalArgumentError, MissingParamError } from '../../utils/errors'
+
+export interface AuthData {
+  gasLimit: number
+  callData: string
```
davidyuk, line 39:
```suggestion
  callData: EncodedData<'cb'>
```

src/contract/ga/helpers.ts
```diff
@@ -1,18 +1,22 @@
...
-  if (typeof authData !== 'object') throw new InvalidAuthDataError('AuthData must be an object')
+export const prepareGaParams = (ins: _Contract) => async (
+  authData: AuthData,
+  authFnName: string): Promise<{authCallData: string, gasLimit: number}> => {
   const gasLimit = authData.gasLimit ?? MAX_AUTH_FUN_GAS
   if (new BigNumber(gasLimit).gt(MAX_AUTH_FUN_GAS)) {
     throw new InvalidAuthDataError(`the maximum gasLimit value for ga authFun is ${MAX_AUTH_FUN_GAS}, got ${gasLimit}`)
   }
-  if (authData.callData) {
+  if (authData?.callData?.length > 0) {
```
davidyuk, line 15: looks too complicated

src/contract/compiler.ts
```diff
@@ -0,0 +1,133 @@
...
+  parameter?: object
+}
+interface ErrorFormatterError {
+  message: string
+  pos: {col: number, line: number}
+  type: string
+  context?: string
+}
+
+export abstract class _ContractCompilerHttp {
```
davidyuk, line 44:
```suggestion
export class _ContractCompilerHttp {
```
it doesn't seems to have abstract methods

src/contract/compiler.ts
```diff
@@ -0,0 +1,133 @@
...
+  compilerApi: CompilerApi
+  compilerVersion: string
+
+  // TODO: replace with constructor after dropping account stamps
+  init ({ compilerUrl, ignoreVersion }: {
+    compilerUrl: string
```
davidyuk, line 50:
```suggestion
    compilerUrl?: string
```

src/contract/compiler.ts
```diff
@@ -0,0 +1,133 @@
...
+            } catch (error) {
+              if (!(error instanceof RestError)) throw error
+              let body: ErrorFormatterError[] | ErrorFormatterBody
+              try {
+                body = JSON.parse(error.response?.bodyAsText ?? '')
+              } catch (e) {
+                throw error
+              }
+              body = body as ErrorFormatterBody
+              error.message = `${new URL(error?.request?.url ?? '').pathname.slice(1)} error`
```
davidyuk, line 86: URL constructor with empty string throws en exception, I wouldn't process error at all if url is not present

src/contract/ga/index.ts
```diff
@@ -0,0 +1,266 @@
...
+export abstract class _GeneralizedAccount extends _Contract {
+  /**
+   * @alias module:@aeternity/aepp-sdk/es/contract/ga
+   * Check if account is GA
+   * @param address - Account address
+   * @returns if account is GA
+   */
+  async isGA (address: string): Promise<boolean> {
+    const { contractId } = await this.getAccount(address)
+    return Boolean(contractId)
```
davidyuk, line 70: can this check be more specific? like
```suggestion
    return contractId != null
```

src/contract/ga/index.ts
```diff
@@ -0,0 +1,266 @@
...
+   * await aeSdk.createGeneralizedAccount(authFnName, authContract, [...authFnArguments]
+   * // Make spend using GA
+   * const callData = 'cb_...' // encoded call data for auth contract
+   * await aeSdk.spend(10000, receiverPub, { authData: { callData } })
+   * // or
+   * await aeSdk.spend(10000, receiverPub, {
+   *   authData: { source: authContract, args: [...authContractArgs] }
+   * }) // sdk will prepare callData itself
+   */
+export const GeneralizedAccount = Contract.compose<_GeneralizedAccount>({
```
davidyuk, line 251: Let's also rewrite this to a set of functions as chain

davidyuk (changes requested):

src/contract/aci.ts
martinkaintas: [`if ((omitUnknown ?? false) && error instanceof MissingEventDefinitionError) return null`
](https://github.com/aeternity/aepp-sdk-js/pull/1511/files#diff-26f6070ca63b8ca32e57bf46ae6c5b2aa050adedb71b463467f4fe1bc183828bR440)returns null

src/contract/aci.ts
martinkaintas: I haven’t worked with autorest or swagger in the past so I don’t think I’m the right person to make the changes required in the `complier.yaml` file. For now I've just extended my ACI interface with the compiled ACI interface.

src/contract/aci.ts
davidyuk: it is removed in `.filter(e => e)`

src/ae/contract.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @alias module:@aeternity/aepp-sdk/es/ae/contract
+ * @category async
+ * @param ids The list of id's to prepend
+ * @param opt Options
+ * @return Signature in hex representation
+ */
+async function delegateSignatureCommon (
+  onAccount: any,
+  ids: Array<EncodedData<EncodingType>> = [],
+  opt: DelegateSigOptions = {}): Promise<string> {
```
davidyuk, line 52:
```suggestion
  opt: DelegateSigOptions = {}
): Promise<string> {
```
just easier to read, also it would be consistent with other code in this repo

src/ae/contract.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Utility method to create a delegate signature for a contract
+ * @alias module:@aeternity/aepp-sdk/es/ae/contract
+ * @category async
+ * @param ids The list of id's to prepend
+ * @param opt Options
+ * @return Signature in hex representation
+ */
+async function delegateSignatureCommon (
+  onAccount: any,
```
davidyuk, line 50: we already have types for accounts, won't we?

src/ae/contract.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @export Contract
+ * @example import { Contract } from '@aeternity/aepp-sdk'
+ */
+
+// @ts-expect-error TODO remove
+import { decode, produceNameId } from '../tx/builder/helpers'
+// @ts-expect-error TODO remove
+import { getNetworkId } from '../node'
+import { EncodedData, EncodingType } from '../utils/encoder'
+import getContractIns from '../contract/aci'
```
davidyuk, line 34:
```suggestion
export { default as getContractInstance } from '../contract/aci'
```

src/ae/index.js
```diff
@@ -193,6 +195,54 @@ const Ae = stampit(NodePool, Tx, AccountResolver, {
...
+          ) {
+            Object.assign(firstArg, {
+              ...instanceOptions,
+              ...firstArg,
+              ...firstArg.onAccount && { onAccount: this._resolveAccount(firstArg.onAccount) }
+            })
+          } else args.unshift(instanceOptions)
+          return handler(...args)
+        }
+      ]
```
davidyuk, line 245: too much code is duplicated here

src/contract/ga/index.ts
```diff
@@ -0,0 +1,171 @@
...
+ * @param args - init arguments
+ * @param options - Options
+ * @returns General Account Object
+ */
+export async function createGeneralizedAccount (
+  { onAccount }: { onAccount: any},
+  authFnName: string,
+  source: string,
+  args: any[] = [],
+  options: object = {}):
```
davidyuk, line 76: Its type should be a combination of options types of functions where it gets passed, like
```suggestion
  options: Parameters<Account['send']>[1] & Parameters<gaAttachTx>[0] = {}):
```

src/contract/ga/index.ts
```diff
@@ -0,0 +1,171 @@
...
+  Promise<Readonly<GAObject>> {
+  const opt = { ...onAccount.Ae.defaults, ...options }
+  const ownerId: string = await onAccount.address(opt)
+  if (await isGA({ onAccount }, ownerId)) throw new IllegalArgumentError(`Account ${ownerId} is already GA`)
+
+  const contract = await onAccount.getContractInstance({ onAccount, source })
```
davidyuk, line 82: `onAccount` shouldn't have `getContractInstance` method, import and call it directly

src/contract/ga/index.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Create a metaTx transaction
+ * @alias module:@aeternity/aepp-sdk/es/contract/ga
+ * @param rawTransaction Inner transaction
+ * @param authData Object with gaMeta params
+ * @param authFnName - Authorization function name
+ * @param options - Options
+ * @returns Transaction string
+ */
+export async function createMetaTx (
+  { onAccount }: { onAccount: any },
```
davidyuk, line 114: This should be moved into the last `options` argument

src/nodeApi.ts
```diff
@@ -55,7 +55,7 @@ export default class extends (Node as unknown as TransformedNode) {
       additionalPolicies: [
         genRequestQueuesPolicy(),
         genCombineGetRequestsPolicy(),
-        genErrorFormatterPolicy((body: ErrorModel) => ` ${body.reason}`)
+        genErrorFormatterPolicy((body: ErrorModel) => ` ${JSON.stringify(body.reason)}`)
```
davidyuk, line 58: `body.reason` should be already defined as string in node api (generated by autorest)

src/ae/index.js
```diff
@@ -193,6 +195,54 @@ const Ae = stampit(NodePool, Tx, AccountResolver, {
...
+    ),
+    ...mapObject(
+      contractMethods,
+      ([name, handler]) => [
+        name,
+        function (...args) {
+          const instanceOptions = {
+            ...this.Ae.defaults,
+            onNode: this.selectedNode.instance,
+            onAccount: this
```
davidyuk, line 207: let's add `onCompiler` here to be able to call compiler methods from `contractMethods`

src/contract/ga/index.ts
martinkaintas: The approach you’re following with chain (putting `onAccount` in the `options` param) doesn’t seem to be working in the case of `ContractMethods` and `GeneralAccountMethods`. Probably because we don’t always pass options to some of those functions or because of the `= {}` initializer. If we put the `onAccount` argument in the `options` param, without passing `options` when calling the function, `onAccount` will end up somewhere else. This is why I put `onAccount` at the top of the parameters. This way it’s always where we expect it to be.

davidyuk: `onAccount` shouldn't be in the first argument, make all arguments before options required then you won't have propblem with detecting the options object

src/contract/aci.ts
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  /**
+  * Deploy contract
+  * @alias module:@aeternity/aepp-sdk/es/contract/aci
+  * @param params Contract init function arguments array
+  * @param options
+  * @returns deploy info
+  */
+  instance.deploy = async (params = [], options: object): Promise<ContractInstance['deployInfo']> => {
```
davidyuk, line 311: options should combine options types of `instance.compile`, `instance.call`, `onAccount.address`, `sendAndProcess`

src/contract/aci.ts
```diff
@@ -0,0 +1,516 @@
...
+
+    const callerId = await onAccount.address(opt).catch((error: any) => {
+      if (opt.callStatic === true) return DRY_RUN_ACCOUNT.pub
+      else throw error
+    })
+    const callData = instance.calldata.encode(instance._name, fn, params)
+
+    let res: any
+    if (opt.callStatic === true) {
+      if (typeof opt.top === 'number') {
```
davidyuk, line 380:
```suggestion
      if (opt.top != null) {
```
we don't have to check its type, just if it present or not

src/contract/aci.ts
```diff
@@ -0,0 +1,516 @@
...
+  * @param options Options
+  * @returns Contract name
+  * @throws {MissingEventDefinitionError}
+  * @throws {AmbiguousEventDefinitionError}
+  */
+  function getContractNameByEvent (
+    address: string,
+    nameHash: BigInt,
+    { contractAddressToName }:
+    {contractAddressToName?: {[key: string]: string}}): string {
```
davidyuk, line 433:
```suggestion
    { contractAddressToName }: { contractAddressToName?: { [key: EncodedData<'ct'>]: string } }
  ): string {
```

src/contract/compiler.ts
```diff
@@ -0,0 +1,117 @@
...
+import { Compiler as CompilerApi } from '../apis/compiler'
+import { UnsupportedVersionError } from '../utils/errors'
+import { genErrorFormatterPolicy } from '../utils/autorest'
+
+interface ErrorFormatterBody {
+  reason?: object
+  info?: object
+  parameter?: object
+}
+interface ErrorFormatterError {
```
davidyuk, line 37:
```suggestion
interface CompilerError {
```
?

src/contract/compiler.ts
```diff
@@ -0,0 +1,117 @@
...
+    })
+    const versionPromise = this.compilerApi.aPIVersion().then(({ apiVersion }) => {
+      this.compilerVersion = apiVersion
+      return apiVersion
+    })
+  }
+}
+
+/**
+    * Contract Compiler Stamp
```
davidyuk, line 98: spacing is broken here

test/integration/node.ts
```diff
@@ -25,7 +25,7 @@ import NodePool from '../../src/node-pool'
 import { NodeNotFoundError } from '../../src/utils/errors'
 
 describe('Node client', function () {
-  let node: Node
+  let node: InstanceType<typeof Node>
```
davidyuk, line 28: looks too complicated, maybe it is because stamps is not removed yet

davidyuk (changes requested):

src/contract/aci.ts
```diff
@@ -0,0 +1,516 @@
...
+}: {
+  onAccount?: Account
+  onCompiler?: _ContractCompilerHttp & {
+    api: any
+  }
+  onNode?: Node
+  source?: string
+  bytecode?: string
+  aci?: Aci
+  contractAddress?: EncodedData<'ak' | 'ct' | 'nm'>
```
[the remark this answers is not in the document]
martinkaintas: We are using `resolveName` (from chain) to get the contract address in aci, which is of type `EncodedData<'ct'>`.
  `resolveName` return type is `EncodedData<'ak' | 'nm'>`.
  Shouldn't it be returning `EncodedData<'ct'>` aswell? I'm proposing these changes to resolveName:

```typescript
  export async function resolveName <Type extends 'ak' | 'ct'> (
  nameOrId: AensName | EncodedData<Type>,
  key: string,
  { verify = true, resolveByNode, onNode }:
  { verify?: boolean, resolveByNode: boolean, onNode: Node }
): Promise<EncodedData< Type | 'nm'>> {
  try {
    decode(nameOrId)
    return nameOrId as EncodedData<Type>
  } catch (error) {}
 //...
}
```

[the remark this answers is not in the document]
davidyuk: sounds good to me, ideally to make `resolveName` return only Promise<EncodedData<Type>> in case `resolveByNode` is true

src/ae/index.js
```diff
@@ -188,7 +188,8 @@ const Ae = stampit(NodePool, Tx, AccountResolver, ContractCompilerHttp, {
             onAccount: this,
             onCompiler: {
               compilerApi: this.compilerApi,
-              compilerVersion: this.compilerVersion
+              compilerVersion: this.compilerVersion,
+              api: this.api
             }
```
davidyuk, line 193:
```suggestion
            onCompiler: this.compilerApi
```
`compilerVersion` not used as I see, `api` provided by `onNode`, so this should be enough

davidyuk (changes requested):

src/contract/aci.ts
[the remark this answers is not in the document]
martinkaintas: I'm proposing to postpone this for now.

davidyuk (approved):

PR #1520, opened 2022-05-19
refactor(tests): migrate remaining tests ts

martinkaintas:


test/integration/accounts.ts
```diff
@@ -31,7 +31,7 @@ import {
 } from '../../src/utils/errors'
 
 describe('Accounts', function () {
-  let aeSdk, aeSdkWithoutAccount
+  let aeSdk: any, aeSdkWithoutAccount: any
```
subhod-i, line 34: @davidyuk Do you think we need a global type for the SDK? we may have better clarity once stampit is removed.
@Kalovelo

test/integration/ga.ts
```diff
@@ -17,12 +17,98 @@
...
+    event: any
+    kind: string
+    state: any
+    type_defs: any[]
+    functions: FunctionACI[]
+  }}
+  externalEncodedAci: any[]
+  interface: string
+}
+interface ContractInstance {
```
subhod-i, line 47: You can reuse these types in the code while removing stamps in contracts section

subhod-i: I mean, I see the same type of definitions in multiple places. Maybe we get to define them while removing stamps.  Not sure if we want to merge this now. we can merge this PR at last.

martinkaintas: Absolutely, these types are declared in the [PR ](https://github.com/aeternity/aepp-sdk-js/blob/7d19119e44fc9c2e135238d2775996ef2e1ee234/src/contract/aci.ts)regarding the RPC module. Once that's merged I will remove them from here and update the PR. Should I rename this to WIP in this case?

subhod-i: I am resolving this. A simple to-do comment will solve this.

test/integration/accounts.ts
subhod-i, line 34: Todo: replace any with type once all stamps are removed and we have a global class/type

test/integration/contract.ts
```diff
@@ -17,12 +17,103 @@
...
-import { BaseAe, getSdk, publicKey, compilerUrl } from './'
+import { BaseAe, getSdk, publicKey, compilerUrl } from '.'
 import {
   IllegalArgumentError, NodeInvocationError, MemoryAccount, generateKeyPair,
   commitmentHash, decode, encode, DRY_RUN_ACCOUNT, messageToHash, salt
+  // @ts-expect-error TODO remove
 } from '../../src'
 
+// TODO remove and import interfaces once contract aci is merged
+interface FunctionACI {
```
subhod-i, line 28: todo: use definitions from the src

subhod-i (approved):
LGTM apart from the pending todos comments.

tsconfig.json
```diff
@@ -13,20 +13,14 @@
...
-      "./node_modules/@types",
-      "./src/typings"
-    ]
+    "typeRoots": ["./node_modules/@types", "./src/typings"]
   },
-  "include": [
-    "src/**/*",
-    "test/integration/*"
-  ]
+  "include": ["src/**/*", "test/integration/**/*", "test/unit/**/*"]
```
davidyuk, line 25: This make ts to add definitions of tests files to the build folder, we need another approach

tsconfig.json
```diff
@@ -13,20 +13,14 @@
...
     "preserveConstEnums": true,
     "declaration": true,
     "downlevelIteration": true,
     "strictNullChecks": true,
     "allowSyntheticDefaultImports": true,
-    "typeRoots": [
-      "./node_modules/@types",
-      "./src/typings"
-    ]
+    "typeRoots": ["./node_modules/@types", "./src/typings"]
```
davidyuk, line 23: avoid changing just formatting

test/utils.ts
```diff
@@ -8,6 +8,6 @@ function randomString (len) {
   return randomString
 }
 
-export function randomName (length) {
+export function randomName (length: number): string {
```
davidyuk, line 11:
```suggestion
export function randomName (length: number): `${string}.chain` {
```

test/unit/memory-account.ts
```diff
@@ -43,23 +43,23 @@ describe('MemoryAccount', function () {
   })
 
   it('Init with secretKey as hex string', async () => {
-    const acc = MemoryAccount({ keypair: testAcc })
-    return acc.address().should.eventually.be.equal(testAcc.publicKey)
+    const acc: _AccountMemory = MemoryAccount({ keypair: testAcc })
+    return await acc.address().should.eventually.be.equal(testAcc.publicKey)
```
davidyuk, line 47: `return` is not necessary then (also change in other places)
```suggestion
    await acc.address().should.eventually.be.equal(testAcc.publicKey)
```

test/unit/memory-account.ts
```diff
@@ -43,23 +43,23 @@ describe('MemoryAccount', function () {
   })
 
   it('Init with secretKey as hex string', async () => {
-    const acc = MemoryAccount({ keypair: testAcc })
-    return acc.address().should.eventually.be.equal(testAcc.publicKey)
+    const acc: _AccountMemory = MemoryAccount({ keypair: testAcc })
```
davidyuk, line 46:
```suggestion
    const acc = MemoryAccount({ keypair: testAcc })
```
didn't we already defined `MemoryAccount` as `_AccountMemory`?

test/unit/memory-account.ts
```diff
@@ -43,23 +43,23 @@ describe('MemoryAccount', function () {
...
+    return await acc.address().should.eventually.be.equal(testAcc.publicKey)
   })
 
   it('Init with secretKey as hex Buffer', async () => {
-    const acc = MemoryAccount({
+    const acc: _AccountMemory = MemoryAccount({
       keypair: {
         publicKey: testAcc.publicKey,
-        secretKey: Buffer.from(testAcc.secretKey, 'hex')
+        secretKey: Buffer.isBuffer(testAcc.secretKey) ? testAcc.secretKey : Buffer.from(testAcc.secretKey, 'hex')
```
davidyuk, line 54: fix https://github.com/aeternity/aepp-sdk-js/issues/1526 instead

test/unit/amount-formatter.ts
```diff
@@ -30,7 +30,7 @@ describe('Amount Formatter', function () {
       [111, AE_AMOUNT_FORMATS.AE, '111000000000000000000'],
       [10012312, AE_AMOUNT_FORMATS.AE, '10012312000000000000000000'],
       [1, AE_AMOUNT_FORMATS.AETTOS, '1']
-    ].forEach(([v, denomination, e]: [number, string, string]) =>
+    ].forEach(([v, denomination, e]: [number, 'ae', string]) =>
```
davidyuk, line 33: `as const` works better in this case (should be already in develop)

test/integration/paying-for.ts
```diff
@@ -66,14 +153,15 @@ describe('Paying for transaction of another account', function () {
...
+    const contract: ContractInstance = await aeSdkNotPayingFee
+      .getContractInstance({ source: contractSource })
+    const { rawTx: contractDeployTx, address } = await contract.deploy([42], {})
     contractAddress = address
     await aeSdk.payForTransaction(contractDeployTx)
     payingContract = await aeSdkNotPayingFee.getContractInstance(
       { source: contractSource, contractAddress }
     )
-    expect((await payingContract.methods.getValue()).decodedResult).to.be.equal(42n)
+    expect((await payingContract.methods.getValue()).decodedResult).to.be.equal(BigInt(42))
```
davidyuk, line 164: why `42n` is not working here?

test/integration/ga.ts
```diff
@@ -59,10 +145,10 @@ describe('Generalized Account', function () {
 
   it('Fail on make GA on already GA', async () => {
     await aeSdk.createGeneralizedAccount('authorize', authContractSource)
-      .should.be.rejectedWith(`Account ${gaAccount.publicKey} is already GA`)
+      .should.be.rejectedWith(`Account ${gaAccount.publicKey.toString()} is already GA`)
```
davidyuk, line 148: isn't it already a string?

test/integration/contract.ts
```diff
@@ -435,7 +525,7 @@ describe('Contract', function () {
       )
       queryExtend.result.returnType.should.be.equal('ok')
       const oracleExtended = await aeSdk.getOracleObject(oracleId)
-      oracleExtended.ttl.should.be.equal(oracle.ttl + 50)
+      oracleExtended.ttl.should.be.equal(parseInt(oracle.ttl) + 50)
```
davidyuk, line 528:
```suggestion
      oracleExtended.ttl.should.be.equal(+oracle.ttl + 50)
```
unary plus is more fragile than `parseInt`, it suits better for tests

test/integration/contract.ts
```diff
@@ -401,20 +491,20 @@ describe('Contract', function () {
...
+    let queryObject: QueryObject
+    let delegationSignature: string
     const queryFee = 500000
     const ttl = { RelativeTTL: [50] }
 
     before(async () => {
       contract = await aeSdk.getContractInstance({ source: oracleContract })
-      await contract.deploy()
-      contractId = contract.deployInfo.address
+      await contract.deploy([], {})
```
davidyuk, line 506: these arguments shouldn't be required

test/integration/contract.ts
```diff
@@ -401,20 +491,20 @@ describe('Contract', function () {
...
+      contractId = contract.deployInfo.address ?? ''
```
davidyuk, line 507:
```suggestion
      contractId = contract.deployInfo.address ?? throw new Error()
```
🤔

package.json
```diff
@@ -17,7 +17,7 @@
     "docs:examples": "node tooling/docs/examples-to-md.js examples/node/*.js",
     "docs:api": "node apidoc.js",
     "lint": "eslint . --ext .ts,.js",
-    "test": "nyc mocha './test/unit/' './test/integration/'",
+    "test": "nyc mocha --require ts-node/register 'test/integration/*' './test/unit/*'",
```
davidyuk, line 20: `ts-node/register` already specified in `.mocharc.js`, also I specified accepted extensions there, so `/*` also not needed

davidyuk (changes requested):

test/integration/contract-aci.ts
```diff
@@ -191,18 +300,18 @@ describe('Contract instance', function () {
...
     expect(deployInfo.txData.gasPrice).to.be.equal(1000000000)
     expect(deployInfo.txData.tx.deposit).to.be.equal(0)
-    expect(testContract.bytecode).to.satisfy(b => b.startsWith('cb_'))
+    expect(testContract.bytecode).to.satisfy((b: string) => b.startsWith('cb_'))
     testContractAddress = deployInfo.address
   })
 
   it('calls', async () => {
-    expect((await testContract.methods.intFn(2)).decodedResult).to.be.equal(2n)
+    expect((await testContract.methods.intFn(2)).decodedResult).to.be.equal(BigInt(2))
```
davidyuk, line 314: I found why are you doing this, it shouldn't be necessary after #1541

davidyuk, line 314: I found why are you doing this, it shouldn't be necessary after #1541

test/integration/ga.ts
martinkaintas: It's `string | Buffer`

davidyuk: `generateKeyPair` definition is fixed now

test/integration/rpc.ts
```diff
@@ -155,8 +172,8 @@ describe('Aepp<->Wallet', function () {
 
     it('Subscribe to address: wallet accept', async () => {
       const accounts = {
-        connected: { [keypair.publicKey]: {} },
-        current: wallet.addresses().reduce((acc, v) => ({ ...acc, [v]: {} }), {})
+        connected: { [keypair.publicKey.toString()]: {} },
```
davidyuk, line 175:
```suggestion
        connected: { [keypair.publicKey]: {} },
```

test/integration/rpc.ts
```diff
@@ -536,12 +560,17 @@ const WindowPostMessageFake = (name) => ({
...
+  // @ts-expect-error
   global.chrome = { runtime: {} }
+  // @ts-expect-error
   global.window = { location: { origin: '//test' }, chrome: global.chrome }
+  // @ts-expect-error
   global.location = { protocol: 'http://test.com' }
   const walletWindow = WindowPostMessageFake('wallet')
   const aeppWindow = WindowPostMessageFake('aepp')
+  if (walletWindow.postMessage == null) throw new InternalError('Expected to not happen, required for TS')
+  if (aeppWindow.postMessage == null) throw new InternalError('Expected to not happen, required for TS')
```
davidyuk, line 573: just fix the return type of `WindowPostMessageFake` instead

davidyuk (changes requested):

test/integration/contract.ts
```diff
@@ -402,19 +494,20 @@ describe('Contract', function () {
...
+    let queryObject: QueryObject
+    let delegationSignature: string
     const queryFee = 500000
     const ttl = { RelativeTTL: [50] }
 
     before(async () => {
       contract = await aeSdk.getContractInstance({ source: oracleContract })
-      await contract.deploy()
+      await contract.deploy([], {})
+      if (contract.deployInfo.address == null) throw new InternalError('Expected to not happen, required for TS')
```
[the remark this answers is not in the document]
martinkaintas: We can add a function like this one, and use it to assert that `deployInfo.address` will be present at a certain scope. But I don't think we can make `deploy` itself do both.

```typescript
 export function assertAddress (address: any): asserts address is EncodedData<'ct'> {
 if (address == null) throw new IllegalArgumentError('Address is required')
}
```

davidyuk (approved):

PR #1522, opened 2022-05-19
refactor(node-pool) drop stamp usage 

### LINKED ISSUE #1521: Drop Stamp usage from node-pool module

Kalovelo:
closes #1521

This PR drops stamp usage on node-pool and also assigns its usage as an entity instead of composition. For example, in order to access `selectedNode`, it shall be done now as `this.nodePool.selectedNode` instead of  `this.selectedNode`

### DIFF
--- a/src/node-pool/index.ts
+++ b/src/node-pool/index.ts
@@ -4,7 +4,6 @@
  * @export NodePool
  * @example import { NodePool } from '@aeternity/aepp-sdk'
  */
-import stampit from '@stamp/it'
 // @ts-expect-error TODO remove
 import { getNetworkId } from '../node'
 import { DisconnectedError, DuplicateNodeError, NodeNotFoundError } from '../utils/errors'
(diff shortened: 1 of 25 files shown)


docs/guides/migration/11.0.0.md
```diff
@@ -94,7 +94,7 @@ console.log(tx.decodedEvents)
 
 ## Following contract instance methods are dropped([#1368](https://github.com/aeternity/aepp-sdk-js/pull/1368))
 * `topBlock`
-    * use `aeSdk.api.getTopHeader()` instead
+    * use `aeSdk.nodePool.api.getTopHeader()` instead
```
subhod-i, line 97:
```suggestion
    * use `aeSdk.api.getTopHeader()` instead
```

This is the migration guide for the older version. The new version will be v12. Can you create a new migration file `12.0.0.md` and add this information there.

src/node-pool/index.ts
```diff
@@ -19,7 +18,7 @@ interface NodeInfo {
   consensusProtocolVersion: string
 }
 
-interface NodePool {
+interface INodePool {
```
subhod-i, line 21: Why do we need an Interface here? @Kalovelo

src/node-pool/index.ts
```diff
@@ -31,33 +30,25 @@ interface NodePool {
...
 }
 
-class _NodePool implements NodePool {
-  readonly api: Node['api']
+export default class NodePool implements INodePool {
   pool: Map<string, Node>
   selectedNode?: Node
 
-  init ({ nodes = [] }: { nodes: Node[] }): void {
+  public get api (): Node['api'] {
```
subhod-i, line 37: Formatting: constructor first, `api` method next down the line?

subhod-i:
Apart from the comments, LGTM.

Kalovelo:
> Apart from the comments, LGTM.

Thank you for your comments, I will update the PR. LGTM seems to be passing without warnings or errors, am I missing something?

docs/guides/migration/11.0.0.md
davidyuk: Actually, we are using `BREAKING CHANGE` commit message

PR #1541, opened 2022-05-27
refactor(NodeApi)!: return BigInt instead of string | number

davidyuk:


tsconfig.json
```diff
@@ -12,8 +12,8 @@
     "outDir": "./es",
     "noImplicitAny": true,
     "module": "es6",
-    "target": "es6",
-    "lib": ["es2019", "dom"],
+    "target": "es2020",
```
davidyuk, line 15: This is needed to support BigInt literals (`10n`), this shouldn't break our dependents because the mjs files in es folder generated by babel (features choosed accodring to browserslist) not by TypeScript

PR #1545, opened 2022-06-01
generateKeyPair: return different type depending on raw value

### LINKED ISSUE #1526: generateKeyPair: return different type depending on `raw` value
So, if `generateKeyPair(true)`, then is should return only Buffers at TypeScript level, otherwise only strings

martinkaintas:
closes #1526


davidyuk (commented):
too many `as` I think I have a simpler solution

davidyuk (approved):
@martinkaintas how do you like it? 😄

martinkaintas:
@davidyuk Looks good! Too many signatures now though 😆

PR #1546, opened 2022-06-02
WIP Generate docs from TSDoc and typescript

martinkaintas:
Notes: 
- doesn’t work with stamps but once stamps are removed it should be fine
- @examples should be wrapped like [this](https://typedoc.org/guides/doccomments/#code-blocks)
- [TypeDoc Documentation](https://typedoc.org/guides/options/)

Usage:  `npm run docs:ts`
To make changes check the `tsdoc.js` file


subhod-i:
I would like to revisit and generate the docs once all PRs are merged. Right now, with stamps, the output looks a bit messy.

davidyuk (approved):

PR #1523, opened 2022-05-20
Stampit removal on contract compiler 

martinkaintas:
closes #1524 
Based on: #1511

PR #1547, opened 2022-06-03
refactor(ae): remove stamps and migrate to TypeScript

### LINKED ISSUE #1517: Migrate ae module to TS

Kalovelo:
based on #1510, closes #1517

As you can see, some stamps have been left (NodePool, AccountResolver, ContractCompilerHttp, AccountMultiple) because they are removed in pending PRs (#1522, #1523)


src/chain.ts
```diff
@@ -51,27 +59,25 @@ export function _getPollInterval (
...
-  * @param tx Transaction to submit
-  * @param options
-  * @param options.onNode Node to use
-  * @param options.onAccount Account to use
-  * @param options.verify Verify transaction before sending
-  * @param options.waitMined Ensure that transaction get into block
-  * @param options.confirm Number of micro blocks that should be mined after tx get included
-  * @return Transaction details
-  */
+ * Submit a signed transaction for mining
```
davidyuk, line 71: would be better to don't change this spacing to reduce diff

src/chain.ts
```diff
@@ -19,7 +19,9 @@
  * Chain module
  * @module @aeternity/aepp-sdk/es/chain
  * @export Chain
- * @example import { Chain } from '@aeternity/aepp-sdk'
+ * @example ```javascript
```
davidyuk, line 22: is `javascript` related to documentation PR, why is it here?

src/chain.ts
```diff
@@ -37,11 +39,17 @@ import {
   Generation, KeyBlock, MicroBlockHeader, NameEntry, SignedTx
 } from './apis/node'
 import { EncodedData } from './utils/encoder'
+import { AensName } from './ae/aens'
 
 export function _getPollInterval (
   type: 'block' | 'microblock',
-  { _expectedMineRate = 180000, _microBlockCycle = 3000, _maxPollInterval = 5000 }
+  options: {
```
davidyuk, line 46: are there a way to keep defaults here?

src/chain.ts
davidyuk, line 71: would be better to don't change this spacing to reduce diff

src/tx/index.ts
```diff
@@ -191,20 +197,22 @@ export async function prepareTxParams (
...
-  },
-  onNode: Node
+    onNode?: Node
+  }
 ): Promise<{
     fee: number | string | BigNumber
     ttl: number
     nonce: number | string | BigNumber
   }> {
+  if (onNode == null) throw new UnexpectedTsError('onNode')
```
davidyuk, line 215: `onNode` should be required if it is used

src/utils/errors.ts
```diff
@@ -15,8 +15,8 @@ export abstract class BaseError extends Error {
 }
 
 export class UnexpectedTsError extends BaseError {
-  constructor (message: string = 'Expected to not happen, required for TS') {
-    super(message)
+  constructor (variable: string = '') {
+    super(`Expected to not happen,${variable} required for TS`)
```
davidyuk, line 19: `Expected to not happen, ${variable} required for TS`

test/integration/chain.js
```diff
@@ -37,20 +37,26 @@ describe('Node Chain', function () {
...
+      .awaitHeight(target)
+      .should.eventually.be.at.least(target)
     await aeSdkWithoutAccount.height().should.eventually.be.at.least(target)
   })
 
   it('Can verify transaction from broadcast error', async () => {
-    const error = await aeSdk.spend(0, publicKey, { fee: 100, verify: false }).catch(e => e)
+    const error = await aeSdk
+      .spend(0, publicKey, { fee: 100, verify: false })
+      .catch((e) => e)
```
davidyuk, line 59: was it necessary to change this?

test/integration/chain.js
```diff
@@ -127,35 +151,51 @@ describe('Node Chain', function () {
...
-  it('ensure transactions mined', () => Promise.all(transactions.map(hash => aeSdkWithoutAccount.poll(hash))))
+  it('ensure transactions mined', () =>
+    Promise.all(transactions.map((hash) => aeSdkWithoutAccount.poll(hash))))
 
   it('multiple contract dry-runs calls at one request', async () => {
     const contract = await aeSdk.getContractInstance({
-      source:
-        'contract Test =\n' +
-        '  entrypoint foo(x : int) = x * 100'
+      source: 'contract Test =\n' + '  entrypoint foo(x : int) = x * 100'
```
davidyuk, line 176: 🤷‍♀️

src/ae/index.ts
```diff
@@ -142,119 +146,117 @@ export class Ae implements IAe {
...
+      ...contractMethods,
+      ...contractGaMethods
+    },
+    ([name, handler]) => [
+      name,
+      function (...args: any[]) {
+        const instanceOptions = {
+          ...this.Ae.defaults,
+          onNode: this.selectedNode.instance,
+          instance: this,
```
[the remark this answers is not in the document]
Kalovelo: The problem is with `onAccount`. Basically, there are cases where `onAccount` has a different value, for example in the transferFunds test: https://github.com/aeternity/aepp-sdk-js/blob/develop/test/integration/accounts.js#L103

onAccount is a reserved word. Therefore, when in need to call `this` functions inside `transferFunds`, error occurs.

A workaround is to use a different keyword, thus `instance`.

Kalovelo:
@davidyuk updated & rebased, only issue is with 2 specific test cases `revoke names` & `transfers names` where I get a unique error: `RestError: v3/transactions/th_2wB71ez... error: Transaction not found`

davidyuk (approved):
I've reverted oracle changes, will merge them separately

PR #1560, opened 2022-06-15
Docs improvements

davidyuk:


martinkaintas:
@davidyuk I'm proposing to wrap `@example` code like [this](https://typedoc.org/guides/doccomments/#code-blocks) so that it gets proper syntax highlighting.

PR #1563, opened 2022-06-15
docs(examples): make it v12 compatible

subhod-i:
Fixed node examples.


subhod-i:
Depends on PR #1567  #1565.

davidyuk (approved):
I think we can name `aeSdk` instances of both `AeSdkWallet` and `AeSdkAepp`, because it is not expected to use any other sdk class at the same time.

docs/contrib/README.md
```diff
@@ -8,9 +8,6 @@ low-level access to it's endpoints, when necessary.
 
 It uses the following Javascript technologies and principles:
 
-- [stampit] provides composable Factories based on the [Stamp
-  Specification]. This is how aepp-sdk approached the [composition over
-  inheritance] principle.
```
davidyuk, line 13: I don't think that we need to keep previously used technologies, someone can check git history if he need

subhod-i:
> 



> I think we can name `aeSdk` instances of both `AeSdkWallet` and `AeSdkAepp`, because it is not expected to use any other sdk class at the same time.

ok then, it is finally **`aeSdk`**.

PR #1567, opened 2022-06-16
fix(wallet): revert to returning complete tx data instead of just hash in rpc wallet signing

### LINKED ISSUE #1561: Web wallet extension(SDK v12) is not compatible with Aepp in examples(SDK v11)
examples/browser/aepp SDK version: 11
web wallet extension SDK version: 12

Building ✅(building with minor changes)
Wallet connection ✅
Contract compile ✅
Spend ✅
Contract deploy ❌

```
Uncaught (in promise) InvalidTxError: Invalid TxObject arguments. Please provide one of { tx: "tx_asdasd23..." } or { type: "spendTx", params: {...} }
    at initTransaction (tx-object.js?436b:130:1)
    at Object.init (tx-object.js?436b:159:1)
    at Stamp (index.js?0cca:34:1)
    at sendAndProcess (aci.js?21f0:148:1)
    at async instance.deploy (aci.js?21f0:234:1)```

### PR DESCRIPTION
**Pull request by subhod-i**
Closes #1561
It's a bit misleading and hard to fix since the earlier releases shipped with schema which returns tx blob in `transactionHash` key. 
This is the fix to make aepps using SDK v11 and below compatible with wallets using SDK v12.

### DIFF
--- a/src/utils/aepp-wallet-communication/rpc/types.ts
+++ b/src/utils/aepp-wallet-communication/rpc/types.ts
@@ -1,3 +1,4 @@
+import { send } from '../../../ae/spend'
 import { EncodedData } from '../../encoder'
 import { METHODS, SUBSCRIPTION_TYPES, WALLET_TYPE } from '../schema'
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/utils/aepp-wallet-communication/rpc/types.ts [hunk 1]
```diff
@@ -1,3 +1,4 @@
+import { send } from '../../..'
```
davidyuk, line 1: I would avoid importing from the `index.ts` to don't have looks like we have a cycle dependencies

src/utils/aepp-wallet-communication/rpc/types.ts
```diff
@@ -43,7 +44,7 @@ export interface WalletApi {
 
   [METHODS.sign]: ((
     p: { tx: EncodedData<'tx'>, onAccount: EncodedData<'ak'>, returnSigned: boolean }
-  ) => Promise<{ transactionHash?: EncodedData<'th'>, signedTransaction?: EncodedData<'tx'> }>)
+  ) => Promise<{ transactionHash?: ReturnType<typeof send>, signedTransaction?: EncodedData<'tx'> }>)
```
davidyuk, line 47: more accurate would be `Awaited<ReturnType<typeof send>>`

davidyuk (approved):
fantastic case 😂

src/utils/aepp-wallet-communication/rpc/types.ts
subhod-i: Good catch. Noted. Auto imported by the IDE it was.

PR #1578, opened 2022-06-22
fix(wallet): make extension connection on demand

subhod-i:
The `browser.runtime.connect()` code in `content-script.js` file opens a connection with the extension wallet on every page resulting in `n` number of RPC clients where n= number of tabs open. Foreach connection wallet keeps on sending its info on `setInterval`.

Solution: Create a runtime connection only when there is a special request from the webpage. The difference between a standard web page and an aepp page is wallet detector. Added a special request `scan`. Overall the connection is established in the `connection-proxy` file only when a webpage sends `scan` request using wallet detector. Existing interfaces work as it is.

Closes https://github.com/aeternity/aepp-sdk-js/issues/1417, https://github.com/aeternity/aepp-sdk-js/issues/907

### DIFF
--- a/src/AeSdkWallet.ts
+++ b/src/AeSdkWallet.ts
@@ -221,6 +221,7 @@ export default class AeSdkWallet extends AeSdk {
           this.onDisconnect(id, disconnectParams); // also related info
         },
         {
+          [METHODS.scan]: async () => this.shareWalletInfo(id),
           [METHODS.closeConnection]: (params) => {
             disconnectParams = params;
             this._disconnectRpcClient(id);
--- a/src/aepp-wallet-communication/connection-proxy.ts
+++ b/src/aepp-wallet-communication/connection-proxy.ts
@@ -16,6 +16,7 @@
  */
 
 import BrowserConnection from './connection/Browser';
+import { METHODS } from './schema';
 
 /**
  * Browser connection proxy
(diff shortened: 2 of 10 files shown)


src/AeSdkWallet.ts
```diff
@@ -221,6 +221,7 @@ export default class AeSdkWallet extends AeSdk {
           this.onDisconnect(id, disconnectParams); // also related info
         },
         {
+          [METHODS.scan]: async () => this.shareWalletInfo(id),
```
subhod-i, line 224: Sharing wallet info on demand. This eliminates `setInterval` loop.

src/aepp-wallet-communication/connection-proxy.ts
```diff
@@ -26,9 +27,12 @@ import BrowserConnection from './connection/Browser';
  * @returns a function to stop proxying
  */
 export default (con1: BrowserConnection, con2: BrowserConnection): () => void => {
-  con1.connect((msg: any) => con2.sendMessage(msg), () => con2.disconnect());
-  con2.connect((msg: any) => con1.sendMessage(msg), () => con1.disconnect());
-
+  con1.connect((msg: any) => {
+    if (!con2.isConnected() && msg.method === METHODS.scan) {
```
subhod-i, line 31: Establish browserRuntime connection only when a page establishes a connection.

src/aepp-wallet-communication/connection/BrowserRuntime.ts
```diff
@@ -24,36 +24,43 @@ import BrowserConnection from './Browser';
...
+    this.port?.disconnect();
   }
 
   connect(
     onMessage: (message: any, origin: string, source: Runtime.Port) => void,
     onDisconnect: () => void,
   ): void {
     super.connect(onMessage, onDisconnect);
-    this.port.onMessage.addListener((message, port) => {
+    this.port = this.port ?? this.setPort();
```
subhod-i, line 50: made `port` optional. connect on demand.

davidyuk (changes requested):
The original goal wasn't formulated properly, we can have extra connection, but they shouldn't appear in the set of rpc clients before aepp is connected. Alternatively, it can be solved by distinglishing connected clients with not connected (I've missed this option before).

More clear way to do don't store extra rpc clients in `wallet._clients` is to extract connection logic from `addRpcClient` and `shareWalletInfo`. And only after aepp called `METHODS.connect` to call `addRpcClient`.

But this refactoring doesn't seems to be required because I found a workaround using `_isRpcClientConnected`

PR #1428, opened 2022-03-21
Fix commitlint

subhod-i:

PR #1579, opened 2022-06-22
Restore codecov and commitlint script

davidyuk:
- codecov was removed in 37bb4afa2b664e58d4c2087249a98760dbb93b2a, I know issues we had there, as a workaround I'm proposing to run all tests at once to send a single report (otherwise it may look buggy), also we need to ensure that we don't need tests that getting executed from time to time to have consistent coverage reports
- commitlint script was removed in #1428 but it is useful to lint commit messages locally


.github/workflows/main.yml
```diff
@@ -30,12 +30,14 @@ jobs:
...
         run: |
           while [[ `curl -s -o /dev/null -w %{http_code} localhost:3013/api` != 200 ]]; do
             sleep 0.2;
           done
-      - run: npm run test:integration
+      - run: npm run test
+      - run: npm run report-coverage
+        if: always()
       - run: docker-compose logs
+        if: always()
```
davidyuk, line 43: The idea was to print container logs even if tests failed, don't know why it was removed

PR #1600, opened 2022-07-07
fix(awaitHeight): allow to wait for arbitrary number of blocks

davidyuk:
`awaitHeight` was accepting the amount of attempts by default equal to 20 that actually should
depend on a difference between current and desired heights. Usually we assume that used node is
being synced therefore any height should be reachable. We are not doing a real-time operating
system, so we don't care about riching a height in a limited time.

Docs related to a similar change in oracles: https://github.com/aeternity/protocol/blob/71cf111ac6b5cc5830999bd130fa0bd753504195/oracles/oracle_transactions.md


thepiwo (approved):
lgtm

PR #1623, opened 2022-07-26
Import specific versions of blakejs and sha.js

davidyuk:
Screenshot 2022-07-26 at 18 25 33 copy
Initial idea was to get rid of `aes-js`, but the best would be to drop `encryptKey`, `decryptKey` functions firstly.


marc0olo:
codecov 🙈

PR #1629, opened 2022-07-27
Export types for use in aepps

loxs:
When scanning for wallets, it's convenient to store the results in some "store" in the frontend and use this to display buttons etc. for the user to decide to which wallet to connect. Currently the types returned by `walletDetector` are not public and can't be used for typing end user code.


davidyuk:
It is intentional 🙂  I'm trying to keep exports clear also to have more space to for refactoring without making breaking changes.

As a workaround you can retrive Wallets type this way
```ts
const wallets: Parameters<Parameters<typeof WalletDetector>[1]>[0]['wallets'] = {};
```

loxs:
Even better, I'll resort to https://zod.dev to sanitize/enforce the types, as they are not only hidden, but also there are errors. For example `Wallet.info` contains `networkId` and the type does not.

Issue #1537, opened 2022-05-26
`NodeApi` should return `EncodedData<...>` instead of a general string


davidyuk:
It is not obvious how to do it in some cases

PR #1678, opened 2022-09-09
fix(channel): getContractCall returned log field may be [Event]

Kalovelo:
apart from `string`, log field may be of contract event type.

```js

const result = await playerChannel.getContractCall({
      caller: responderConfig.responderId,
      contract: contractAddress,
      round: playerChannel.round(),
    });

/*
    {
      callerId: 'ak_XUChq1DGgu4Wj2f1x14cUN4u4jmMk19veJ8GYfv6qjPPVLjx2',
      callerNonce: 5,
      contractId: 'ct_2cCWyFBHuKDqadojkx9AWm4k992oHTUQFXFgcwHbop7MjFzMmH',
      gasPrice: 1,
      gasUsed: 24695,
      height: 5,
      log: [
        [Object: null prototype] {
          address: 'ct_2cCWyFBHuKDqadojkx9AWm4k992oHTUQFXFgcwHbop7MjFzMmH',
          data: 'cb_YWtfWFVDaHExREdndTRXajJmMXgxNGNVTjR1NGptTWsxOXZlSjhHWWZ2NnFqUFBWTGp4Mnxha19DM05FdlNhZGpKOHNDZEJCbVN3TDZpeVBpVldNaEFwVmlaV3JzUllYb2UxVEE3NlNrNRkyVg==',
          topics: [Array]
        },
        [Object: null prototype] {
          address: 'ct_2cCWyFBHuKDqadojkx9AWm4k992oHTUQFXFgcwHbop7MjFzMmH',
          data: 'cb_cGFwZXI79bMK',
          topics: [Array]
        }
      ],
      returnType: 'ok',
      returnValue: 'cb_r4IAAQA/aHG2bw=='
    }
*/
```


src/channel/index.ts
```diff
@@ -908,7 +909,7 @@ export default class Channel {
       gasPrice: number | BigNumber;
       gasUsed: number | BigNumber;
       height: number;
-      log: string;
+      log: string | [Event];
```
davidyuk, line 912: `[Event]` as I know is a single-size tuple with Event, but there can be multiple events as in your example
also, I think `string` was added there by a mistake, so I'm replacing it with `Event[]`

davidyuk: Also, `callerId`, `callerNonce`, `contractId` is missed here, but they a defined in a `ContractCallObject`, so I would use it instead. The only problem is that `ContractCallObject` uses `string` instead of `xx_`-prefixed strings, it has to be fixed in Node Api generator 🙁

src/contract/Contract.ts
```diff
@@ -81,7 +81,7 @@ interface Aci extends BaseAci {
   externalEncodedAci: any[];
 }
 
-interface Event {
+export interface Event {
```
davidyuk, line 84: actually, `Event` is here because of https://github.com/aeternity/aepp-sdk-js/issues/1537, it is already defined in autorest-generated api

PR #1680, opened 2022-09-10
Refactor channels

davidyuk:
I didn't find time to finish this refactoring, but that done so far should also work


Kalovelo (approved):
Can't wait for the next release! :rocket: 

Since this refactoring is not 100% complete, noting here that there should be these options on channel options as well:
```
 minimumDepthStrategy: 'plain' | 'txFee';
 minimumDepth: number;
 fee: BigNumber | number;
```

Also, I see that some _private_ (underscore prefixed) fields for example `channel._fsm` are available to be used - is this intended or they're not supposed to be available (for example by stating them as `private`) ?

davidyuk:
> `channel._fsm`

This is temporary, I would use `protected` for them and the other properties when inline `handlers.ts` and `internal.ts` to classes. Currently, it can't be done because it won't be accessible outside the class.

PR #1708, opened 2022-11-22
feat(channel): add helper to get contract address created by other peer

### LINKED ISSUE #1619: A channel helper to get address of contract created by other peer
Contract address is calculated by contract owner address and the current round. Contract owner is available while signing `channels.sign.update_ack`, but the new round is applied after `channels.update` message.

Possible options:
- add contract address to `update` key of `channels.sign.update_ack` (this would be a bit early because new round is not confirmed yet)
- add some `onContract` callback that will be fired when other peer created a contract in channel
- track channel state (all balances), and add new contract there silently

Kalovelo:
closes #1619 

This PR adds a stateless support for created contracts on channel.
Peers can listen on `onNewContract` events and also access created contract addresses through `channel.contracts()`.

Due to the fact that the node API does not provide way to fetch deployed contract addresses, the implementation is stateless which means that for example in (browser) cases, where users refresh their window and reconnects, they will lose access to given array of contract addresses. Given this, `channel.contracts()` might be a good discussion point whether if it should be kept or removed

### DIFF
--- a/src/channel/Base.ts
+++ b/src/channel/Base.ts
@@ -33,6 +33,7 @@ import {
   ChannelAction,
   ChannelStatus,
   ChannelFsm,
+  ChannelMessage,
 } from './internal';
 import { ChannelError } from '../utils/errors';
 import { Encoded } from '../utils/encoder';
@@ -78,7 +79,7 @@ export default class Channel {
 
   _fsmId?: Encoded.Bytearray;
 
-  _messageQueue: object[] = [];
+  _messageQueue: ChannelMessage[] = [];
 
   _isMessageQueueLocked = false;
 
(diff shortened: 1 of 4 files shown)


davidyuk:
> Due to the fact that the node API does not provide way to fetch deployed contract addresses

This is not actually true, I find out that Channel.state() methods return `trees` property that holds the whole state encoded into `ss_`-prefixed string. I've improved `unpackTx` function to support this
https://github.com/aeternity/aepp-sdk-js/pull/1737/files#diff-476caac91ea1ab94e08e9a57d42a7eb3cb7888e6151d74a89677770b2bbe746bR242-R303

Now it is possible to get addresses of all created contracts by
```js
Object.keys(unpackTx((await channel.state()).trees).contracts)
```
later I will make `channel.state()` to return already unpacked structures. So, I'm going to revert `contracts` method and its state, but keep `newContract` event.

davidyuk (approved):

PR #1709, opened 2022-11-22
docs: add guide how to transfer fraction of funds

### LINKED ISSUE #1609: add guide how to send a fraction of funds (including 100%) from an account to another
this is important to know for people that wanna empty their account and transfer all of their funds to another account

martinkaintas:
closes #1609

### DIFF
--- a/examples/node/transfer-ae.mjs
+++ b/examples/node/transfer-ae.mjs
@@ -70,6 +70,8 @@ console.log(`Balance of ${recipient} (before): ${balanceBefore} aettos`);
 // Calling the `spend` function will create, sign and broadcast a `SpendTx` to the network.
 const tx = await aeSdk.spend(amount, recipient);
 console.log('Transaction mined', tx);
+// Alternatively, you can use [transferFunds](https://docs.aeternity.com/aepp-sdk-js/latest/api/functions/transferFunds.html)
+// method to transfer a fraction of your AE to another account.
 
 // ## 6. Get AE balance of recipient (after transfer)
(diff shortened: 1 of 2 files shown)


examples/node/transfer-ae.mjs
```diff
@@ -80,3 +80,10 @@ console.log(`Balance of ${recipient} (after): ${balanceAfter} aettos`);
 //  - If the recipient is the same account as the sender (default of the script if no arguments
 //    provided) the balance will be lower after transfer because a transaction `fee` has been
 //    paid to the miners.
+
+// ## 7. Transfer a fraction of your AE to another account
+// Instead of `spend` function you can use the `transferFunds` function
+// to transfer a fraction of your AE to another account
+const fraction = 1;
```
marc0olo, line 87: maybe provide an additional hint that "1" will transfer 100% and "0.5" would transfer 50% to eliminate any confusion :)

davidyuk (commented):
I'm fine to merge this, but I think that `transferFunds` is an advanced feature, and would be better to create a separate guide for this keeping `spend` example focused/simple

marc0olo:
agree, could be in a separate section and more highlighted 🤔

davidyuk:
I've checked guides and actually, all of them cover bigger scopes of features. I think the best is to link this example with `transferFunds` in API reference. So, a newcomer would know that we have API reference. Also, I'm adding an example to API reference to make it as clear as possible:


Screenshot 2023-01-27 at 11 10 26

PR #1718, opened 2022-12-01
docs(aens): remove usages of removed `createAensDelegationSignature`

### LINKED ISSUE #1461: Drop `createAensDelegationSignature`, `createOracleDelegationSignature`
https://github.com/aeternity/aepp-sdk-js/pull/1439#pullrequestreview-921407481

davidyuk:
Was missed in 24699f15896be0984fc8396979148fa97288dab4
closes #1461

### DIFF
--- a/docs/guides/aens.md
+++ b/docs/guides/aens.md
@@ -368,17 +368,16 @@ This functionality could for example be used to build an AENS marketplace.
 
 const aeSdk = new AeSdk({ ... }) // init the SDK instance with AeSdk class
 
-// initialize contract
-const contract = await aeSdk.initializeContract({ address: 'ct_asd2ks...', aci })
+const contractAddress = 'ct_asd2ks...'
 // AENS name
 const name = 'example.chain'
 // Sign with a specific account
 const onAccount = aeSdk.address
 
 // this signature will allow the contract to perform a pre-claim on your behalf
-const preClaimSig = await contract.$createDelegationSignature()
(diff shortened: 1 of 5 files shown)


marc0olo:
🤔 can't this still be used? it is implemented in https://github.com/aeternity/aepp-sophia-examples/blob/master/test/aensDelegation.js#L65

will there be a breaking change in the new SDK version? does that mean a contract instance is needed to create the delegation signature?

davidyuk:
> will there be a breaking change in the new SDK version?

Yep, it would be sdk@13

> does that mean a contract instance is needed to create the delegation signature?

Yes, from my perspective these signature delegation is highly connected to contract functionality

marc0olo:
> Yes, from my perspective these signature delegation is highly connected to contract functionality

disagree here. I only need the contract address to create the delegation signature. it's independent of the logic of any contract. so now every user who wants to provide a delegation signature needs to somehow create a contract instance just to be able to provide that signature. is that correct?

if that's the case I would ask you to rollback this change please

marc0olo:
maybe we can support both ways, not sure. but to me it's not linked to the actual contract logic

davidyuk:
I don't have a strong opinion here, so the creation of a delegation signature would be a method of AeSdk again.

marc0olo:
> I don't have a strong opinion here, so the creation of a delegation signature would be a method of AeSdk again.

thanks @davidyuk ! :)

PR #1720, opened 2022-12-06
fix: compatibility with @vue/cli@4

davidyuk:
based on https://github.com/subhod-i/aepp-sdk-js/commit/a7d816822c045cedbecb4204e38bc06d4b109986
I don't think that SDK needs `@babel/plugin-proposal-private-property-in-object` plugin because it doesn't uses `#property in object` (https://babeljs.io/docs/en/babel-plugin-proposal-private-property-in-object)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,8 +2,13 @@ module.exports = {
   presets: [
     ['@babel/preset-env', {
       include: [
+        // compatibility with create-react-app@4
         '@babel/plugin-proposal-nullish-coalescing-operator',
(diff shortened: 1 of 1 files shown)


subhod-i:
@davidyuk isn't it a private property in here https://github.com/aeternity/aepp-sdk-js/blob/bf916a7034375cad3447ad071a51df553b9d7c58/src/aepp-wallet-communication/rpc/RpcClient.ts#L42

subhod-i (approved):

davidyuk:
> isn't it a private property in here

it is, but I'm not using it in combination with `in` operator

PR #1736, opened 2023-01-17
Simplify transaction builder 3

### LINKED ISSUE #1727: Types discrepancy in transaction parameters(buildTx and unpackTx)
### buildTx
The options object passed to the `buildtTx` method even in the tests resembles the same error. 

```
Argument of type '{ senderId: `ak_${string}`; recipientId: `ak_${string}`; amount: number; payload: string; }' is not assignable to parameter of type 'Omit<Omit<Omit<Partial<Pick<{ payload: string | undefined; }, "payload">> & Omit<{ payload: string | undefined; }, "payload">, "tag" | "VSN"> & { ...; }, "nonce" | ... 3 more ... | "abiVersion"> & { ...; }, "onNode"> & { ...; }'.
  Object literal may only specify known properties, and 'senderId' does not exist in type 'Omit<Omit<Omit<Partial<Pick<{ payload: string | undefined; }, "payload">> & Omit<{ payload: string | undefined; }, "payload">, "tag" | "VSN"> & { ...; }, "nonce" | ... 3 more ... | "abiVersion"> & { ...; }, "onNode"> & { ...; }'.ts(2345)
 ```
 Data passed in here is not assignable to the defined types https://github.com/aeternity/aepp-sdk-js/blob/2c992b9cfb056adcb5468b4306fee02a2e2e0063/test/integration/rpc.ts#L291
 
 ### unpackTx
Property 'fee' does not exist on type 'RawTxObject<Partial<Pick<{ payload: string | undefined; }, "payload">> & Omit<{ payload: string | undefined; }, "payload">>'.
 
 ```
 const { fee } = unpackTx(spendTx, Tag.SpendTx).tx;
 ```
 The above code indeed has `fee` data in the return type at runtime but the type checker throws an error at the compile time.
  

 Expected behavior: compatible types as expected.

### PR DESCRIPTION
**Pull request by davidyuk**
based on the work done in #1730 
closes #1727

### CODE REVIEW

File: src/index-browser.ts [hunk 1]
```diff
@@ -19,11 +19,13 @@ export * from './chain';
 export * from './utils/crypto';
 export * from './utils/keystore';
 export * from './utils/bytes';
-export * from './tx';
 export * from './tx/builder';
 export * from './tx/builder/helpers';
 export * from './tx/builder/constants';
-export * from './tx/builder/schema';
+// TODO: move to constants
```
subhod-i, line 25: @davidyuk you may have missed this

src/tx/builder/field-types/ct-version.ts
```diff
@@ -35,14 +36,33 @@ export function getProtocolDetails(
...
 
+  async prepare(
+    value: CtVersion | undefined,
+    params: {},
+    // TODO: { consensusProtocolVersion: ConsensusProtocolVersion } | { onNode: Node } | {}
+    options: { consensusProtocolVersion?: ConsensusProtocolVersion; onNode?: Node },
+  ): Promise<CtVersion | undefined> {
+    if (value != null) return value;
+    if (options.consensusProtocolVersion != null) return undefined;
+    if (Object.keys(ConsensusProtocolVersion).length === 2) return undefined;
```
subhod-i, line 56: Could '2' be in the constants?

src/tx/builder/field-types/deposit.ts
```diff
@@ -8,14 +8,16 @@ export default {
   /**
    * @param value - Deposit value in string format. Should be equal to '0'.
    * @param options - Options
+   * @param parameters - Parameters
    * @returns Deposit value Buffer.
    */
   serialize(
     value: Int | undefined,
     options: Parameters<typeof coinAmount['serialize']>[1],
+    parameters: Parameters<typeof coinAmount['serialize']>[2],
```
subhod-i, line 17: Maybe we could have a type alias with a readable name for such type builds. `Parameters<typeof coinAmount['serialize']>[2]`

src/index-browser.ts
davidyuk: nope, it is definitely to be fixed later

src/tx/builder/field-types/ct-version.ts
davidyuk: `2` means that `ConsensusProtocolVersion` have a single protocol version, e.g.
```
{
  'Iris': 5,
  '5': 'Iris',
}
```
In this case we don't need to find out the current protocol because it won't be compatible with sdk if it is different what we know.
It can be extracted only as `KeysCountInEnumWithASlingleElement`, but I would leave it as a TS quirck.

src/tx/builder/field-types/deposit.ts
davidyuk: This is quite verbose, but the base idea is just to copy the signature of the inherited method. I'm doing it frequently with different objects and their parts, so I'm not sure that adding intermediate types would be helpful in all cases, but I'm doing so in the most popular, like this:
https://github.com/aeternity/aepp-sdk-js/blob/6b8888b7d8bb168f2f3dd0f9737e88dce4889ffe/src/tx/index.ts#L34-L35

PR #1740, opened 2023-02-01
chore(deps): mkdocs version update

marc0olo:


marc0olo:
nothing urgent. but as I updated again in protocol repo I thought it is good to update everywhere

PR #1744, opened 2023-02-07
feat(node): exponential retry of requests on failure

### LINKED ISSUE #1630: Add load balancer policy to node pipeline
It should do "exponential retries with short timeout". Be optional, but enabled by default (to be compatible with mainnet.aeternity.io).

Dincho's suggestion:
> out of thin air I'd say: 3 retries starting with 100ms delay (then go x2)
that should total to ~600ms
if you start with 200ms, it would total to 1400ms
might work better
or you can target a total timeout of X then decide on what to start

davidyuk:
closes #1630 
found issues https://github.com/aeternity/aeternity/pull/4084

Attempt | Intermediate | Delay
-- | -- | --
0 | 0 | 0
1 | 0.33 | 55.55
2 | 0.66 | 222.22
3 | 1 | 500
  | Delay sum | 777.77

This is a bit different than Dincho suggested, I would reduce the amount of retries and make it truly exponential 🙂

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -3,7 +3,7 @@ import BigNumber from 'bignumber.js';
 import { OperationArguments, OperationSpec } from '@azure/core-client';
 import {
   genRequestQueuesPolicy, genCombineGetRequestsPolicy, genErrorFormatterPolicy,
-  genVersionCheckPolicy,
+  genVersionCheckPolicy, genRetryOnFailurePolicy,
 } from './utils/autorest';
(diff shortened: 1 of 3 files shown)


davidyuk:
@dincho this should work well in case of 404, but seems I can't do anything with an outdated value (like nonce) on this level

dincho:
I don't think it needs to handle "outdated" nonce.

I would expect this to be configurable? Like enabling, number of retries and exponential backoff configuration

dincho (approved):

PR #1746, opened 2023-02-07
chore: remove useless autorest parameters

davidyuk:


tooling/autorest/compiler.yaml
```diff
@@ -24,7 +24,6 @@ use-extension:
 input-file: compiler-swagger.yaml
 output-folder: ../../src/apis/compiler
 source-code-folder-path: .
-client-side-validation: true
```
davidyuk, line 27: initially, I wanted to disable this option but seems this is not working, even though it was implemented before https://github.com/Azure/autorest.typescript/issues/243

PR #1779, opened 2023-03-27
feat(aepp): call onNetworkChange after connection

davidyuk:

### DIFF
--- a/src/AeSdkAepp.ts
+++ b/src/AeSdkAepp.ts
@@ -92,8 +92,7 @@ export default class AeSdkAepp extends AeSdkBase {
    */
   async connectToWallet(
     connection: BrowserConnection,
-    { connectNode = false, name = 'wallet-node', select = false }:
-    { connectNode?: boolean; name?: string; select?: boolean } = {},
+    { connectNode = false, name = 'wallet-node' }: { connectNode?: boolean; name?: string } = {},
   ): Promise<WalletInfo> {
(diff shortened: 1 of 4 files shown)


src/AeSdkAepp.ts
```diff
@@ -92,11 +92,19 @@ export default class AeSdkAepp extends AeSdkBase {
...
+  ): Promise<WalletInfo & { node?: NodeRpc }> {
     if (this.rpcClient != null) throw new AlreadyConnectedError('You are already connected to wallet');
     let disconnectParams: any;
+
+    const updateNetwork = (params: Network): void => {
+      if (connectNode) {
+        if (params.node == null) throw new RpcConnectionError('Missing URLs of the Node');
+        this.addNode(name, new Node(params.node.url), true);
+      }
+      this.onNetworkChange(params);
```
subhod-i, line 105: Minor refactoring
```suggestion
  if (!connectNode) {
    return;
  }

  if (!params.node?.url) {
    throw new RpcConnectionError('Missing URLs of the Node');
  }

  this.addNode(name, new Node(params.node.url), true);
  this.onNetworkChange(params);
```

subhod-i (approved):

src/AeSdkAepp.ts
davidyuk: `onNetworkChange` needs to be called regardless `connectNode`, and it can't be moved above because I expect it to be fired after network switching complete. Casting to boolean using `!` is not allowed by linter. I'm accepting `node?.url` part.

PR #1782, opened 2023-03-30
fix(chain): reduce /2 number of blocks to wait before throwing timeout

davidyuk:
Currently if transaction can't be mined but present in mempool, sdk by default would wait for 10 blocks before sdk would fail. Considering it takes 3 minutes to mine a generation, this would fail after 30 minutes, that is extremely long time to wait. I propose to reduce this 2 times (to 15 minutes). Thought it can be even smaller, 2 blocks?

### DIFF
--- a/src/chain.ts
+++ b/src/chain.ts
@@ -76,7 +76,7 @@ export async function getHeight({ onNode }: { onNode: Node }): Promise<number> {
 export async function poll(
   th: Encoded.TxHash,
   {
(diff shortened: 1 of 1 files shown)


loxs:
Depends on what we are doing, but 15 minutes is also a huge amount of time for a web app to be waiting for. Still better than 30 minutes.

loxs (approved):

thepiwo (approved):

PR #1805, opened 2023-05-02
Implement MiddlewareWebsocket

davidyuk:


thepiwo (approved):
probably fine, didn't check in detail

PR #1825, opened 2023-05-17
fix: `onAccount` option in AeSdkMethods

davidyuk:
Before it was failing in `getValueOrErrorProxy(() => this._resolveAccount())` because of no argument provided.

I'm going to try AeSdkMethods instead of other classes because it has less state and simpler in general.

### DIFF
--- a/src/AeSdkMethods.ts
+++ b/src/AeSdkMethods.ts
@@ -11,25 +11,29 @@ import Node from './Node';
 import { TxParamsAsync } from './tx/builder/schema.generated';
 import AccountBase from './account/Base';
 import { Encoded } from './utils/encoder';
-import { ArgumentError, NotImplementedError, TypeError } from './utils/errors';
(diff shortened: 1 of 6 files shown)


src/AeSdkMethods.ts
```diff
@@ -150,15 +135,13 @@ Object.assign(AeSdkMethods.prototype, mapObject<Function, Function>(
...
-      };
+      args[args.length - 1] = this._getOptions(options);
       return handler(...args);
     },
   ],
 ));
 
-export default AeSdkMethods as new (options?: ConstructorParameters<typeof AeSdkMethods>[0]) =>
-AeSdkMethods & AeSdkMethodsTransformed;
+type AeSdkMethodsTyped = AeSdkMethods & AeSdkMethodsTransformed;
```
davidyuk, line 144: based on https://stackoverflow.com/a/64661625/6176994

PR #1837, opened 2023-05-30
provide types in exports field of package.json

davidyuk:
Fixes compatibility with @vue/tsconfig

https://github.com/vuejs/tsconfig#migrating-from-typescript--50
https://nodejs.org/api/packages.html#community-conditions-definitions


thepiwo (approved):
lgtm

PR #1843, opened 2023-06-18
Support signing typed data

### LINKED ISSUE #1781: Implement ability to sign typed structured data
An analog of EIP-712, https://eips.ethereum.org/EIPS/eip-712
Would be useful in signing auth data in GA contracts.

call data can encode decode arbitrary structures https://github.com/aeternity/aepp-calldata-js/blob/master/src/api/ContractByteArrayEncoder.js

davidyuk:
closes #1781

found issues https://github.com/aeternity/aepp-calldata-js/issues/215 https://github.com/aeternity/aepp-calldata-js/issues/216 https://github.com/aeternity/aesophia/issues/461


src/utils/typed-data.ts
```diff
@@ -0,0 +1,90 @@
...
+  } as const;
+  return hash(decode(encodeFateValue(domain, domainAci)));
+}
+
+export function hashTypedData(
+  data: Encoded.ContractBytearray,
+  aci: AciValue,
+  domain: Domain,
+): Buffer {
+  return hash(concatBuffers([hashDomain(domain), hashJson(aci), hash(decode(data))]));
```
davidyuk, line 89: Should prefix the calculated hash with a magic number? I see that Ethereum developers care to have not overlapped signing payloads for transactions and other messages, but I don't understand the reasoning. For example,

> The initial 0x19 byte is intended to ensure that the signed_data is not valid RLP

https://eips.ethereum.org/EIPS/eip-191

src/utils/typed-data.ts
```diff
@@ -0,0 +1,90 @@
+// @ts-expect-error see https://github.com/aeternity/aepp-calldata-js/issues/216
+import ContractByteArrayEncoder from '@aeternity/aepp-calldata/src/ContractByteArrayEncoder';
```
davidyuk, line 2: need to add `.js` extension and disable eslint, otherwise it won't work in esm

src/utils/typed-data.ts
hanssv: Yes I think there should be a prefix, though maybe not 0x19 since that was chosen because `Ethereum Signed Message:` happens to be 0x19 bytes long...

davidyuk: I propose to define 0x1a00 prefix, with the second byte as a signature type (0x00 for typed data). E.g. 0x1a61 would be a prefix of a signed message the same as in EIP-191. This won't overlap with transaction signatures because they are prefixed with network id that can't contain ASCII control characters (0x1a, substitute).

davidyuk: I propose to define 0x1a00 prefix, with the second byte as a signature type (0x00 for typed data). E.g. 0x1a61 would be a prefix of a signed message the same as in EIP-191. This won't overlap with transaction signatures because they are prefixed with network id that can't contain ASCII control characters (0x1a, substitute).

Issue #1845, opened 2023-06-21
Add methods to generate aci by Sophia interface to ContractBase


davidyuk:
aesophia_cli has [--no-code option to allow ACI creation from stub/partial contract](https://github.com/aeternity/aesophia_cli/pull/79)

PR #1847, opened 2023-06-23
feat(compiler): add `generateAci`, `generateAciBySourceCode`

davidyuk:
closes https://github.com/aeternity/aepp-sdk-js/issues/1845

blocked by:
- release of aesophia_cli
- https://github.com/aeternity/aesophia_http/pull/121

### DIFF
--- a/src/contract/compiler/Base.ts
+++ b/src/contract/compiler/Base.ts
@@ -56,6 +56,32 @@ export default abstract class CompilerBase {
   }>;
 
   /**
+   * Generate contract's ACI by contract's path
+   * Available only in Node.js
+   * @param path - Path to contract source code
+   * @returns ACI
+   */
+  abstract generateAci(path: string): Promise<Aci>;
+
+  /**
+   * Generate contract's ACI by contract's source code
+   * @param sourceCode - Contract source code as string
+   * @param fileSystem - A map of contract filename to the corresponding contract source code to
+   * include into the main contract
+   * @example
+   * ```js
+   * {
+   *   'library.aes': 'namespace TestLib =\n  function sum(x: int, y: int) : int = x + y'
+   * }
+   * ```
+   * @returns ACI
+   */
+  abstract generateAciBySourceCode(
+    sourceCode: string,
+    fileSystem?: Record<string, string>,
+  ): Promise<Aci>;
+
+  /**
    * Verify that a contract bytecode is the result of compiling the given source code
    * Available only in Node.js
    * @param bytecode - Contract bytecode to verify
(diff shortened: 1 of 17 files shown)


src/contract/compiler/Http.ts
```diff
@@ -81,6 +81,25 @@ export default class CompilerHttp extends CompilerBase {
     throw new NotImplementedError('File system access, use CompilerHttpNode instead');
   }
 
+  async generateAciBySourceCode(
+    sourceCode: string,
+    fileSystem?: Record<string, string>,
+  ): Promise<Aci> {
+    try {
+      return await this.api.generateACI({ code: sourceCode, options: { fileSystem } });
```
thepiwo, line 89: return can also be without await

src/contract/compiler/Base.ts
```diff
@@ -55,6 +55,32 @@ export default abstract class CompilerBase {
...
+   * ```js
+   * {
+   *   'library.aes': 'namespace TestLib =\n  function sum(x: int, y: int) : int = x + y'
+   * }
+   * ```
+   * @returns ACI
+   */
+  abstract generateAciBySourceCode(
+    sourceCode: string,
+    fileSystem?: Record<string, string>,
```
thepiwo, line 81: isn't there a type alias for this?

src/contract/compiler/Cli.ts
```diff
@@ -99,6 +99,28 @@ export default class CompilerCli extends CompilerBase {
...
+    }
+  }
+
+  async generateAciBySourceCode(
+    sourceCode: string,
+    fileSystem?: Record<string, string>,
+  ): Promise<Aci> {
+    const tmp = await CompilerCli.#saveContractToTmpDir(sourceCode, fileSystem);
+    try {
+      return await this.generateAci(tmp);
```
thepiwo, line 118: also return await?

thepiwo (approved):

src/contract/compiler/Base.ts
davidyuk: I know only the original type `{ [k: string]: string }` 🤷‍♀️ in types generated from swagger file it is defined as `fileSystem?: Record<string, unknown>`

src/contract/compiler/Http.ts
davidyuk: it is because of try/catch
```js
async function foo() {
  try {
    return await Promise.reject(new Error('test'));
  } catch (error) {
    console.log('error handled');
  }
}

foo().catch(() => console.log('error thrown'));
```
error would be handled differently depending on `await` in this case

thepiwo: ok, sounds good

src/contract/compiler/Cli.ts
thepiwo: ok, sounds good

src/contract/compiler/Base.ts
thepiwo: ok :)

PR #1870, opened 2023-07-28
docs(compiler): explain `compilerPath` argument and related fixes

davidyuk:

### DIFF
--- a/src/contract/compiler/Cli.ts
+++ b/src/contract/compiler/Cli.ts
@@ -17,14 +17,19 @@ const getPackagePath = (): string => {
 };
 
 /**
- * A wrapper around aesophia_cli, available only in Node.js
- * Assumes that `escript` is available in PATH.
+ * A wrapper around aesophia_cli, available only in Node.js.
+ * Requires Erlang installed, assumes that `escript` is available in PATH.
  */
 export default class CompilerCli extends CompilerBase {
   #path: string;
 
   #ensureCompatibleVersion = Promise.resolve();
 
+  /**
+   * @param compilerPath - A path to aesophia_cli binary, by default uses the integrated one
+   * @param options - Options
+   * @param options.ignoreVersion - Don't ensure that the compiler is supported
+   */
   constructor(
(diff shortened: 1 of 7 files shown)


src/contract/compiler/Cli.ts
```diff
@@ -17,14 +17,19 @@ const getPackagePath = (): string => {
...
+ * A wrapper around aesophia_cli, available only in Node.js.
+ * Requires Erlang installed, assumes that `escript` is available in PATH.
  */
 export default class CompilerCli extends CompilerBase {
   #path: string;
 
   #ensureCompatibleVersion = Promise.resolve();
 
+  /**
+   * @param compilerPath - A path to aesophia_cli binary, by default uses the integrated one
```
marc0olo, line 29: "integrated one" means that the CLI is actually shipped along with the SDK. do I understand correctly?

marc0olo: I assume that windows is not supported here using the CLI compiler, right? 🤔

marc0olo:
it would also be great to have this explained in the contract guide here:
- https://github.com/aeternity/aepp-sdk-js/blob/develop/docs/guides/contracts.md

few things I would love to see added/adjusted:

1. explain the difference between CLI compiler and http compiler
    - also, provide a note that the "hosted" compiler will be decommissioned and tell the user to run their own local http compiler if needed or to install the erlang dependency to use the shipped CLI
    - also provide a note that windows doesn't support the CLI compiler (if that's the case)
1. show usage of compiler methods directly that show how to:
    - generate & use bytecode & ACI (seems currently missing)
1. refer to or explain the utility method to get the filesystem map

src/contract/compiler/Cli.ts
marc0olo: that's cool. wouldn't have expected the CLI binary of the compiler to work with windows

davidyuk: it is an Erlang VM binary

Issue #1786, opened 2023-04-06
A way to estimate transaction fee to get a reasonable mining time in a high blockchain load


davidyuk:

subhod-i:
I am thinking if  'calculateFee` can return a floor and ceiling value. What do you think?

```javascript
{
    minFee: 00000160000000
    maxFee: 00000200000000
}
```

this would help aepps developer to implement a fee slider too for the users.

davidyuk:
Ideally it should be a mapping of expected time to fee

Like in Ethereum before POS https://ethgasstation.info:
- less then 2 minutes => set gas price to 53
- less then 5 minutes => set gas price to 14
- less then 30 minutes => set gas price to 8

davidyuk:
Let's assume that all transactions have the same type and size. Each mined transaction has a fee. We are looking for a minimum fee to pay to get our transaction mined in a specific time interval (like the next 5 minutes). We can't predict it by previously mined transactions for sure because demand can vary significantly due to external reasons.

As a greedy solution I propose to use the minumum fee paid by somebody in the last N minutes. Using that fee, our transaction would be **probably** mined in the next N minutes. Would be efficient to calculate it using middleware, but for now I can do it on the sdk (client) side.

PR #1884, opened 2023-08-23
feat(tx-builder): get actual gas price from node

davidyuk:
closes https://github.com/aeternity/aepp-sdk-js/issues/1786

To protect from funds loss in case of a bug/misusage, I propose to add an upper limit for the gas price substituted by default. Known issues:

### A single active client would compete with himself

If somebody posts transactions from a single client with high blockchain utilization, the gas price will increase for no reason. This is a special use case, I'm adding a note to `batch-requests.md`.

### Overall solution doesn't protect from mining time fluctuation

With a low utilization, it is expected to get a transaction mined within 3 seconds. Currently, it is cheap to increase utilization at any moment. The current solution would reflect utilization change in a minute. So, a user may experience arbitrary changes of mining time.

Screenshot 2024-03-12 at 14 57 57

I've wrote a script to create several clients pushing transactions one by one and to visualize that gathered data. On the screenshot, I've added about 5 dynamic price clients and removed all of them with an interval from the middle to the end. It looks +/- as expected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   node:
-    image: aeternity/aeternity:master@sha256:824d40d60fcf4a07d805bba6b6880e182331a971bb8e4793d45d66eaa5e568fa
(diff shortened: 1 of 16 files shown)


dincho (approved):
Better than static. The test included in this PR would be very helpful in future too.

PR #1890, opened 2023-08-27
chore: replace node with middleware

davidyuk:


davidyuk:
Middleware is not usually up to date with the resent node changes, also it doesn't support the devmode completely. It is fine to run node and mdw at the same time as it is now.

PR #1946, opened 2024-02-15
Add NodeGateway retrying nonce requests if response outdated

davidyuk:

We are retrying requests multiple times if the response status code is 404 or so. It is made because a public API gateway has a load balancer, and some nodes behind it can be a bit unsynced producing outdated results. The same issue with nonce tracking, node can return an outdated nonce, but it is much harder to detect. A wrong nonce completely blocks tx mining (see https://github.com/aeternity/aeternity/issues/4173).

To solve this problem I've implemented a class inherited from Node. This class extracts nonces from submitted transactions and if the nonce returned by node looks outdated then it retries the request the same way if it would be 404. Sdk tests became much more stable using this on testnet.

A general solution would be https://github.com/aeternity/aepp-sdk-js/issues/291.


dincho (approved):
I did only functional review. This is formal approval of the new strategy not the codebase/implementation.

If I understand it correctly this changes are making sure that the gateway/node/backend nonce is in sync with the SDK instance (kind of/partially). In general I don’t see possible regressions (other than extra error that must be probably handled by the apps), and only eventual improvement.

dincho:
I think the default 8 retries are a bit too much, while currently there are no RPS limits we'll implement it at some point (as protection measure) which can cause issues.

In general I don't think the SDK should care about of the "gateway" concepts and should tread it as ordinal node/API backend, however the user/developer should be able to choose the proposed nonce strategy for any backend.

Also hardcoding the endpoints also sounds like a bad idea, yes we should keep them as defaults, but the users should be able to change it?

dincho:
> Also hardcoding the endpoints also sounds like a bad idea, yes we should keep them as defaults, but the users should be able to change it?

Okay I was confused (didn't fully read) these lines: https://github.com/aeternity/aepp-sdk-js/pull/1946/files#diff-9f5cadd5139250f5eef2e796974c1db0f422f7adb4541f5b5b356a26c54711afR43

It still means one would get the waring if they don't use our gateways? A bit annoying I guess as there are multiple environments, user nodes, hyperchains etc.

davidyuk:
> It still means one would get the waring if they don't use our gateways?

It is a hint to use `NodeGateway` in case the old `Node` class is used with our API gateways.

Issue #1859, opened 2023-07-17
Consider setting default ttl to 5 blocks


davidyuk:
To avoid account blocking for a long time if the transaction is invalid. Also, it would be consistent with https://github.com/aeternity/aepp-sdk-js/pull/1782. Polling would be more accurate because if tx wasn't mined in 5 blocks then it won't be mined for sure.

PR #1956, opened 2024-03-04
Set ttl to height + 3 for transactions built internally

davidyuk:

How sdk would work after this PR:
- transactions built using `buildTx`, `AeSdk:buildTx`, `buildTxAsync` will have ttl set to 0 (as before)
- transactions built while executing sdk's methods (like `aeSdk.spend(...)`) would have ttl set to `~height + 3`, with the ability to override this value though global and method's options `{ ttl: 0 }`
- poll of transaction with ttl set to 0 by default will wait for 5 blocks before rejecting (as before)
- poll of transaction with ttl set to not 0 will wait till transaction would be removed from mempool (the same as `ttl` expires)

Doing this way sdk would ensure that the transaction won't get stuck in mempool, so it won't affect next-nonce calculation. Also, this transaction won't get mined after polling is rejected, so it can be safely retried.

closes https://github.com/aeternity/aepp-sdk-js/issues/1859

### DIFF
--- a/src/aens.ts
+++ b/src/aens.ts
@@ -45,6 +45,7 @@ export async function aensRevoke(
   options: AensRevokeOptions,
 ): ReturnType<typeof sendTransaction> {
   const nameRevokeTx = await buildTxAsync({
+    _isInternalBuild: true,
     ...options,
     tag: Tag.NameRevokeTx,
     nameId: name,
@@ -96,6 +97,7 @@ export async function aensUpdate(
   }
 
   const nameUpdateTx = await buildTxAsync({
+    _isInternalBuild: true,
     ...options,
     tag: Tag.NameUpdateTx,
     version: hasRawPointers ? 2 : 1,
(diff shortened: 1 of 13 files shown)


dincho:
Sounds good to me, why `buildTx` will keep using ttl: 0? Backward compatibility? Do you plan changing it in future?

davidyuk:
Because prepared tx can be signed as inner transaction and be paid by somebody else and I'm not sure that 3 blocks is enough then. Also, I'm planning to set nonce to 0 for internal transactions to override this nonce as needed before sending.

So, the idea is that `buildTx` should generate correct and most-cases-compatible transactions, and `buildTx` in "internal" mode should generate transactions fits the best to other sdk's methods.

thepiwo (approved):

davidyuk:
> when the chain is more congested 3 blocks might not be sufficient for tx to be included

I plan to implement tx retries (with re-signing) on sdk side for cases like this.

PR #1970, opened 2024-03-28
feat(aepp,wallet): allow raw data sign

davidyuk:
The ability to sign raw data implemented as a fallback in case the payload that aepp needs to sign is not compatible with existing methods.

Wallet needs to warn user if it can't parse the data to sign. Or wallet can disable this method, to work only with a known payloads.


thepiwo (approved):
lgtm

Issue #1806, opened 2023-05-02
Check compatibility with Deno


davidyuk:
Metin tried to use sdk in Deno and said that it is not working. In general there a lot of entusiasts in blockchain development, so it makes sense to support edge cases like this.

PR #1997, opened 2024-06-18
chore!: remove keystore implementation (AEX-3)

### LINKED ISSUE #1799: Use higher memory requirements with argon2
https://github.com/aeternity/aepp-sdk-js/blob/79a2c5a2b49d11d4d3dfb11d51b0d01e1296aa71/src/utils/keystore.ts#L63

Consider increasing the memory used in the kdf.

Ref:
- https://blog.dataparty.xyz/blog/wtf-is-a-kdf/

davidyuk:
Keystore was implemented according to [AEX-3].

I can't find a good package implementing argon2 (used in keystore) for both nodejs and browser:
- [argon2] — good for nodejs, not working in browser
- [argon2-browser] — poorly maintained, latest release 3 years ago
- [@node-rs/argon2] — wasm version takes 200kb instead 20kb in argon2-browser, requires a [workaround for npm]
- [libsodium-sumo] — was used by sdk some years ago, increases bundle size significantly because it supports multiple algorithms

WASM version can be used in nodejs, but it is noticeably slower.

Currently, sdk uses [@aeternity/argon2] which is a combination of `argon2` and `argon2-browser`. `argon2` updates frequently because they adds new binaries, so our library should be updated consistently. The last time I had issues with this setup was when I tried to run sdk in Deno #1806.

AEX-3 didn't get wide support, the apps I know that using it are [aepp-cli-js] and seem [waellet]. The AEX-3 use case in general is unclear,
- mnemonics phrases are mostly used in user-facing wallets
- on the server side storing keystore and password is basically the same as storing a secret key (new fancy format implemented in #1991)

The only use case is a kind of wallet that requires a password provided by the user to interact with. But most wallets encrypt keys by themself, and if they need to exchange secrets the mnemonic phrase is used.

In case of dropping AEX-3 in sdk, I would copy it to `aepp-cli-js` repo to don't make breaking changes in cli.

Considering the expenses of maintaining this compared to usefulness, I propose to drop the keystore support.

Alternatives to dropping AEX-3:
- replace argon2 with a different crypto primitive like something from [Web Crypto API]
- ask a developer to provide argon2 implementation by himself
- maintain @aeternity/argon2 (update deps, write a readme, add tests, ensure that browser and nodejs works the same)

[libsodium-sumo]: https://www.npmjs.com/package/libsodium-sumo
[Web Crypto API]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API
[@aeternity/argon2]: https://github.com/aeternity/argon2-js
[workaround for npm]: https://github.com/napi-rs/node-rs/issues/810
[argon2]: https://www.npmjs.com/package/argon2
[argon2-browser]: https://www.npmjs.com/package/argon2-browser
[AEX-3]: https://github.com/aeternity/AEXs/blob/master/AEXS/aex-3.md
[@node-rs/argon2]: https://www.npmjs.com/package/@node-rs/argon2
[aepp-cli-js]: https://github.com/aeternity/aepp-cli-js
[waellet]: https://forum.aeternity.com/t/aex-3-secure-storage-format/3220/2

closes #1799

BREAKING CHANGE: `recover`, `dump` removed (AEX-3 keystore implementation)
Copy the removed implementation to your project or add a previous sdk as a separate dependency
```json
  "dependencies": {
    "@aeternity/aepp-sdk": "^14.0.0",
    "@aeternity/aepp-sdk-13": "npm:@aeternity/aepp-sdk@^13.3.2"
  }
```


thepiwo:
Is it being used by anyone as far as we know?

dincho:
So this is not about keys deriving from menmonic but only for key encryption in rest?

davidyuk:
> So this is not about keys deriving from mnemonic but only for key encryption in rest?

Yep. By AEX-3, it also can be used to encrypt mnemonic, but sdk is not aware of that. Key derivation by mnemonic defined in [AEX-10](https://github.com/aeternity/AEXs/blob/master/AEXS/aex-10.md).

dincho:
> > So this is not about keys deriving from mnemonic but only for key encryption in rest?
> 
> Yep. By AEX-3, it also can be used to encrypt mnemonic, but sdk is not aware of that. Key derivation by mnemonic defined in [AEX-10](https://github.com/aeternity/AEXs/blob/master/AEXS/aex-10.md).

Then I think it even  _must_ be removed then. As long the tools support common format mnemonic +deriving algo interoperability should be fine

davidyuk:
I've also asked about this in Slack, but nobody responded. So it looks like I can merge this.

Issue #1672, opened 2022-09-04
Expose secret key of `MemoryAccount`


davidyuk:
For security reasons (as I understand), secret keys of MemoryAccounts stored separately and inaccessible from outside of memory account module.
https://github.com/aeternity/aepp-sdk-js/blob/66568bebba9a3bdc3344e13fdcd4da867da11404/src/account/Memory.ts#L29

`sign` method is not protected anyhow, so if malicious actor have direct access to `MemoryAccount` instance, then he can sign arbitrary data (like `SpendTx`). Also he can stole the account by making it generalized (similar to knowing the private key).

Protecting the private key doesn't make significant benifits because of the above, but it decreases usability of recently entroduced `generate` method. Currently `generate` can't be used if needed to persist the generated account because of no access to secret key. I propose to add `secretKey` getter to allow using `MemoryAccount` instead of `generateKeyPair`, and to deprecate the last.

If necessary, the secret key hidding can be implemented on user side by inheriting from `MemoryAccount`, like
```js
class MemoryAccountProtected extends MemoryAccount {
  get secretKey() {
    throw new Error('Access to private key is forbidden.');
  }
}
```

nikita-fuchs:
Shouldn't we rather protect `sign` also? making basically everything unprotected by default would open up risks when using ae in the backend, and we would have to tell people that the safe way of doing things is not the default MemoryAccount, but you need to use something different.

Why would we need a getter for the secret key in memory account, as it should be known to the dev anyway? It's used for generating the memory account after all.

davidyuk:
> Shouldn't we rather protect sign also?

the idea is that developer should build something on the top of MemoryAccount, like an UI that ask user to confirm signing, or a route hook that ensures that request is legitimate

> Why would we need a getter for the secret key in memory account

As I wrote above, otherwise the account generated by `MemoryAccount.generate()` can't be persisted, it reduces usability of `generate` method.

PR #1998, opened 2024-07-03
Add mnemonic factory, remove hd wallet functions

davidyuk:
closes https://github.com/aeternity/aepp-sdk-js/issues/1672

If necessary, you can make a version of MemoryAccount not exposing a secret key by
```ts
class MemoryAccountProtected extends MemoryAccount {
  override secretKey: Encoded.AccountSecretKey = 'sk_hidden';
}
```


src/account/Memory.ts
```diff
@@ -31,28 +27,29 @@ export function getBufferToSign(
 export default class AccountMemory extends AccountBase {
   override readonly address: Encoded.AccountAddress;
 
+  readonly #secretKeyDecoded: Uint8Array;
```
davidyuk, line 30: Private properties the same secure as using WeakMap

> In prototypal inheritance, its behavior may be emulated with WeakMap objects

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

thepiwo (commented):
isn't that a breaking change? can we document the migration path in the changelog?

davidyuk:
@thepiwo  As before I will build migration guide by commit descriptions starting with "BREAKING CHANGE: ...". But for some methods I won't have a good guide because I think they was not popular, like hd wallet methods.

thepiwo (approved):

PR #2031, opened 2024-11-18
Update low/high level api docs

davidyuk:

### DIFF
--- a/docs/guides/low-vs-high-usage.md
+++ b/docs/guides/low-vs-high-usage.md
@@ -2,109 +2,37 @@
 
 ## Interactions
 
-> "There are two approaches, purist and high-level."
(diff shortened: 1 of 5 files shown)


dincho (commented):
Not sure why you're using absolute links to external services/docs while you can use relative to the repo. You plan to update the docs on each release version?

Also the external files examples doesn't make much sense as it's short and can be inlined for better readability flow, otherwise one have to click around.
The same question/comment applies here for absolute urls.

davidyuk:
> can be inlined for better readability flow

they are actually inlined by GitHub
Screenshot 2024-11-19 at 23 05 54
and I also make it work in MkDocs. The problem is that there are only absolute links supported as I understand.
Screenshot 2024-11-19 at 23 11 28
Also, it requires to specify lines, so I'm referencing the commit because if I add a line, it won't show in the docs.
Keeping example code in separate files also allow testing them, to ensure that they are still functional.

PR #2061, opened 2025-03-29
Update Docs Links for Conract Events

adasauce:
Just some broken links I found in the SDK docs as I was reading through.


davidyuk (approved):
thank you!

PR #2159, opened 2026-03-23
fix: set default TTL to 0 (no TTL)

davidyuk:
initially implemented in https://github.com/aeternity/aepp-sdk-js/pull/1956


src/tx/builder/field-types/ttl.ts
```diff
@@ -3,6 +3,9 @@ import Node from '../../../Node.js';
 import { ArgumentError } from '../../../utils/errors.js';
 import { _getPollInterval, getHeight } from '../../../chain.js';
 
+// TODO: decrease to a reasonable value after the mainnet mining issue is solved
+const DEFAULT_INTERNAL_RELATIVE_TTL = (30 * 24 * 60) / 3;
```
CedrikNikita: I do not know what would be the right ttl default time to be honest.

CedrikNikita (approved):

CedrikNikita (approved):

PR #2185, opened 2026-08-06
feat(tx-builder): get consensus parameters from node

CedrikNikita:


venimus:
Reviewed against the current `aeternity/aeternity` `master` (cross-checked `tooling/autorest/node.yaml`, `src/tx/builder/protocol-parameters.ts`, and the tests against the node's actual, merged OpenAPI spec). The design here is solid: bounding node-supplied values against a plausibility ceiling so a misconfigured/malicious node can't inflate fees, the TTL cache with a shorter fallback interval, the `rebuildUnpackedTx`/`serializeAsIsParam` mechanism to avoid re-validating an already-priced transaction — all good. `MIN_GAS_PRICE`/`MAX_AUTH_FUN_GAS` kept as deprecated rather than removed is good backward compat too.

## Blocking issue: the node API shape has moved since this was written

`tooling/autorest/node.yaml` pins the spec to commit `613cd4b0687b4fcd0fb8a7deb48bd80b6f90e52e` — an in-progress commit from aeternity's `feature/add-protocol-parameters-endpoint` branch (already flagged in this PR's own `TODO: RELEASE BLOCKER` comment). That branch has since merged as `aeternity/aeternity@8a4e86b8f` (#4692), but **the response shape changed between the pinned commit and the actual merge**:

- `node_settings` was nested inside `/protocol-parameters` at the pinned commit; on the merged spec it's a **separate `/node-settings` endpoint** (`NodeSettings` schema), not nested anymore.
- `block_gas_limit` moved from being **per-protocol** (`ProtocolConsensusParameters.block_gas_limit`) to living only in `NodeSettings`.
- `callable_vm_versions` was removed from `ProtocolConsensusParameters` entirely (low impact — only referenced in this PR's own test fixtures, not in `src/`).

`convertProtocolParameters()` currently does:
```ts
const { currentProtocolVersion, protocols, nodeSettings } = response;
...
blockGasLimit: protocol.blockGasLimit,
maxAuthFunGas: nodeSettings.maxAuthFunGas,
```
Both `response.nodeSettings` and `protocol.blockGasLimit` no longer exist in the merged shape. This isn't caught by CI because the generated client (`src/apis/node`, gitignored/regenerated) is self-consistent with this PR's own stale spec pin — CI never compares against the real, now-merged node schema.

## What's needed to merge
1. Re-pin `tooling/autorest/node.yaml` to a real tag/commit that includes the merged `#4692` shape (once aeternity cuts a release containing it), not an in-progress branch commit.
2. Update `convertProtocolParameters` to request/merge **two** endpoints (`getProtocolParameters` + `getNodeSettings`) instead of reading `nodeSettings` off one response.
3. Source `blockGasLimit` from the node-settings response instead of per-protocol.
4. Update `test/unit/node.ts`, `test/integration/node.ts`, and `test/unit/protocol-parameters.ts` fixtures accordingly (and drop the now-nonexistent `callable_vm_versions` test fixture field).

Everything else in the PR looks ready; this is the one thing standing between this and being mergeable.

venimus (approved):

PR #2186, opened 2026-08-06
feat(tx-builder): support Arcus consensus protocol

CedrikNikita:


src/Node.ts
```diff
@@ -53,7 +53,7 @@ export default class Node extends NodeApi {
     super(url, {
       allowInsecureConnection: true,
       additionalPolicies: [
-        genVersionCheckPolicy('node', getVersion, '7.1.0', '9.0.0', ignoreVersion),
+        genVersionCheckPolicy('node', getVersion, '7.1.0', '8.0.0', ignoreVersion),
```
CedrikNikita, line 56: There is no plan on raising the node version.

venimus (approved):

